PR #5, opened 2020-06-18
Introduce conventions for event and data messages

jskeet:


jskeet:
@yolocs PTAL

jskeet:
Merging - this is all still pending a full review, but I think it gets us in a better place.

Issue #6, opened 2020-06-18
Storage Object schema discrepancies


jskeet:
Three sources of truth to consider:

- [storage_resources.proto](https://github.com/googleapis/googleapis/blob/master/google/storage/v1/storage_resources.proto)
- [Knative Storage schema](https://github.com/google/knative-gcp/blob/master/schemas/storage/schema.json)
- What we see in GCF Storage events - [sample 1](https://github.com/GoogleCloudPlatform/functions-framework-dotnet/blob/master/src/Google.Cloud.Functions.Framework.Tests/GcfEvents/storage.json), [sample 2](https://github.com/google/knative-gcp/tree/master/docs/examples/cloudstoragesource#verify)

Aspects to consider:

- `event_based_hold` is currently a `google.protobuf.BoolValue` - can we just make it `bool`?
- Should we make `crc32c` a `string`? That would match `md5_hash`
- `selfLink`, `mediaLink` and `kind` are all present in the GCF event, but not in our current schema. Add them?
- `owner` and `acl` are present in the Knative schema, but we don't know if they'll ever actually be seen.

cc @nachocano

jskeet:
Note: if we make changes, this may make translating between representations trickier. We need to think about this from an infrastructure perspective as well as consumer. This shouldn't stop us from doing things; it just makes it a slightly bigger deal.

jskeet:
I've added `self_link` and `media_link`.

We believe that in the current system, `owner` and `acl` will never be included in event data, so it would be useful not to include them (particularly as they mean introducing more message types). We still need to consider the other aspects here.

jskeet:
Confirmed that `event_based_hold` can be converted to `bool`, which is the last change we need in this issue.

PR #7, opened 2020-06-22
Add fields which are in existing JSON, but not the API proto

jskeet:
The field numbers are intended to be high enough to avoid conflicts
with new API fields.

Note that "kind" is not included here, as it's not expected to be
useful.

Part of addressing #6


jskeet:
cc @yolocs

jskeet:
Merging this now to make the schema review later today easier.

PR #11, opened 2020-06-23
Add the "wrapper" around the PubSub message

jskeet:
This corresponds to the JSON received for push notifications. In GCF
events, there is no indication of the subscription; only the message
is received as the event data. This means we cannot convert from the
GCF format to the CloudEvent format.

cc @yolocs @hrasadi


hrasadi:
In the worse case scenario where we need to perform the format upcasting (which I believe should not happen at all), we still have an engineering path forward to make the missing field available.

PR #12, opened 2020-06-23
Remove "envelope" protos

jskeet:
(First commit is from #11)

@yolocs @hrasadi

### DIFF
--- a/proto/README.md
+++ b/proto/README.md
@@ -10,19 +10,10 @@ created to make it easier to consume events.
 
 ## Relationship between protobuf messages and CloudEvent types
 
-There are two protobuf messages that are relevant for any given CloudEvent
-type:
-
-- The "envelope" message (e.g. `ObjectFinalizedEvent`)
-- The "data" message (e.g. `StorageObjectData`)
-
-There is a 1:1 relationship between CloudEvent types and envelope
-messages, but a many:1 relationship between CloudEvent types and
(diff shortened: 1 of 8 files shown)


yolocs:
I’m more leaning towards discussing this before removing. We obviously lack a place to put event types (markdown doesn’t appear to be the right place to me).

yolocs:
Sorry I missed that you moved types to data messages. I don’t quite understand the reasoning behind. Could you explain?

jskeet:
>  I don’t quite understand the reasoning behind. Could you explain?

Specifically so we would still have that information available, because it's useful information. This was what was suggested in the meeting yesterday.

I can see three options:

- Do more work to make the envelope messages more useful (e.g. with more concrete use cases, supported by libraries)
- Remove the envelope messages without moving the event types (remove the annotation itself for the moment)
- Go half way, as we have in this PR.

jskeet:
> The CE dataschema which events should provide shoould point directly to the proto seen in the data field.

Yes, and that was always the expectation. The messages that are being removed here were always "nice to have for some use cases" rather than "the core messages every developer will need to use" - the latter being the `*Data` messages.

jskeet:
Closing this PR as we have an alternative approach now: keep the protos themselves, but use proto file naming to allow selective generation. (Consumer libraries probably only need the data protos for now.)

PR #19, opened 2020-06-25
Rename scheduler data message

yolocs:
Renamed to `JobExecutedData`. This would be consistent with other data messages naming.


yolocs:
@jskeet should we get this in?

jskeet:
Yes, let's go for it. We won't be getting any more data, so this is probably as good as it gets :)

PR #20, opened 2020-06-30
Event registry generation tool

jskeet:
This is very drafty - it's just the *kind* of thing we can do.

Sample output:

```text
google.cloud.audit.log.v1.written:
  Event message: google.events.cloud.audit.v1.AuditLogWrittenEvent
  Data message: google.events.cloud.audit.v1.AuditLogData
google.cloud.firestore.document.v1.created:
  Event message: google.events.cloud.firestore.v1.DocumentCreatedEvent
  Data message: google.events.cloud.firestore.v1.DocumentEventData
google.cloud.firestore.document.v1.updated:
  Event message: google.events.cloud.firestore.v1.DocumentUpdatedEvent
  Data message: google.events.cloud.firestore.v1.DocumentEventData
google.cloud.firestore.document.v1.deleted:
  Event message: google.events.cloud.firestore.v1.DocumentDeletedEvent
  Data message: google.events.cloud.firestore.v1.DocumentEventData
google.cloud.firestore.document.v1.written:
  Event message: google.events.cloud.firestore.v1.DocumentWrittenEvent
  Data message: google.events.cloud.firestore.v1.DocumentEventData
google.cloud.pubsub.topic.v1.messagePublished:
  Event message: google.events.cloud.pubsub.v1.MessagePublishedEvent
  Data message: google.events.cloud.pubsub.v1.MessagePublishedData
google.cloud.scheduler.job.v1.executed:
  Event message: google.events.cloud.scheduler.v1.JobExecutedEvent
  Data message: google.events.cloud.scheduler.v1.SchedulerData
google.cloud.storage.object.v1.finalized:
  Event message: google.events.cloud.storage.v1.ObjectFinalizedEvent
  Data message: google.events.cloud.storage.v1.StorageObjectData
google.cloud.storage.object.v1.archived:
  Event message: google.events.cloud.storage.v1.ObjectArchivedEvent
  Data message: google.events.cloud.storage.v1.StorageObjectData
google.cloud.storage.object.v1.deleted:
  Event message: google.events.cloud.storage.v1.ObjectDeletedEvent
  Data message: google.events.cloud.storage.v1.StorageObjectData
google.cloud.storage.object.v1.metadataUpdated:
  Event message: google.events.cloud.storage.v1.ObjectMetadataUpdatedEvent
  Data message: google.events.cloud.storage.v1.StorageObjectData
google.cloud.firebase.database.document.v1.created:
  Event message: google.events.firebase.v1.DocumentCreated
[...]

PR #21, opened 2020-07-01
Use a smaller extension field number for the cloud_event_type

jskeet:
While the field number here *should* be valid, it at least causes
problems in C#. I'll investigate whether this is a Google.Protobuf bug
separately, but it does no harm to use a smaller number here.


jskeet:
(This is effectively the first commit from #20, but as a stand-alone PR.)

jskeet:
cc @yolocs

PR #22, opened 2020-07-01
Update Firestore event data proto to match internal field ordering

jskeet:
This doesn't affect the JSON representation at all, but we might as
well match internal proto field numbers where possible - it may have
advantages in the future if we want to deliver proto binary
representations of events.


jskeet:
cc @yolocs

PR #23, opened 2020-07-01
Fix Firebase event types

jskeet:
(These are deliberately not in a "cloud" directory, and so "cloud" shouldn't be in the event type.)

@grant: Apologies for the deluge of tiny PRs today!


jskeet:
cc @yolocs

PR #41, opened 2020-08-06
Add and update Firebase events (in separate packages)

jskeet:
This is like #38, but with different Firebase products in different packages.

I expect the RemoteConfig event type to require a bit more discussion - see the specific commit for more info.

Cc:
- @eburley
- @yolocs
- @mbleigh
- @hrasadi
- @fredzqm


jskeet:
For some reason the Database events haven't made it into the README. I'm looking into that.

jskeet:
> For some reason the Database events haven't made it into the README. I'm looking into that.

Fixed. I'd got "events.data" instead of "events.proto". Doh :)

proto/google/events/firebase/analytics/v1/events.proto
```diff
@@ -0,0 +1,31 @@
...
+
+package google.events.firebase.analytics.v1;
+
+import "google/events/firebase/analytics/v1/data.proto";
+import "google/events/cloudevent.proto";
+
+option csharp_namespace = "Google.Events.Protobuf.Firebase.Analytics.V1";
+
+// The CloudEvent raised when a Firebase Analytics log is written.
+message AnalyticsLogWritten {
```
yolocs, line 25: At one point I think we want to have "Event" suffix for messages in `events.proto`?

README.md
```diff
@@ -29,7 +29,10 @@ This repository contains definitions for the following CloudEvents:
 |[google.events.cloud.pubsub.v1](proto/google/events/cloud/pubsub/v1)|google.cloud.pubsub.topic.v1.messagePublished|MessagePublishedData|
 |[google.events.cloud.scheduler.v1](proto/google/events/cloud/scheduler/v1)|google.cloud.scheduler.job.v1.executed|SchedulerJobData|
 |[google.events.cloud.storage.v1](proto/google/events/cloud/storage/v1)|google.cloud.storage.object.v1.archived<br/>google.cloud.storage.object.v1.deleted<br/>google.cloud.storage.object.v1.finalized<br/>google.cloud.storage.object.v1.metadataUpdated|StorageObjectData|
-|[google.events.firebase.v1](proto/google/events/firebase/v1)|google.firebase.database.document.v1.created<br/>google.firebase.database.document.v1.deleted<br/>google.firebase.database.document.v1.updated<br/>google.firebase.database.document.v1.written|DocumentEventData|
+|[google.events.firebase.analytics.v1](proto/google/events/firebase/analytics/v1)|google.firebase.analytics.log.v1.written|AnalyticsLogData|
```
grant, line 32: Can you separate the table rows for these separate products?

proto/google/events/firebase/analytics/v1/events.proto
jskeet: Yes - will fix.

README.md
jskeet: Not sure what you mean each product (proto package) has its own table row.

jskeet:
I'd really rather not make the message names too long. `FirebaseDatabaseReferenceEventData` is 35 characters long. So a variable declaration and initialization like this:

```csharp
FirebaseDatabaseReferenceEventData x = new FirebaseDatabaseReferenceEventData();
```

... is already 80 characters (a line limit in some systems) without *any* indentation and with a terrible variable name.

The information is already in the namespace - I don't think we need to qualify it any further. (The reason we did for `StorageObjectData` is that "object" is so massively overloaded - ditto for `Message` really for PubSub.)

Happy to talk more about this, but 35 character type names would really annoy me personally as a user.

README.md
grant: nvm. Looks like the GitHub md preview doesn't add zebra stripes for previews. Looks good when checking out locally.

grant:
> I'd really rather not make the message names too long. `FirebaseDatabaseReferenceEventData` is 35 characters long. So a variable declaration and initialization like this:
> 
> ```cs
> FirebaseDatabaseReferenceEventData x = new FirebaseDatabaseReferenceEventData();
> ```
> 
> ... is already 80 characters (a line limit in some systems) without _any_ indentation and with a terrible variable name.
> 
> The information is already in the namespace - I don't think we need to qualify it any further. (The reason we did for `StorageObjectData` is that "object" is so massively overloaded - ditto for `Message` really for PubSub.)
> 
> Happy to talk more about this, but 35 character type names would really annoy me personally as a user.

Sounds good. Good to understand where we are going on the scale.

jskeet:
Okay - I'm going to merge these now, because it'll then be easier for others to review later. This shouldn't be seen as making anything cast in stone though. (Basically @mbleigh I hear you, and I'm not holding you to this :)

Issue #39, opened 2020-08-03
Add Example JSON payloads per Event Type


grant:
We should add example JSON payloads per event type for easy understanding of event payloads.

This will also make it easy for languages that don't have libraries to implement their own clients.

CC: @michaelawyu

jskeet:
They could also act as canonical resources for test purposes.

grant:
Here is an example for [`MessagePublishedData`](https://github.com/googleapis/google-cloudevents/blob/master/proto/google/events/cloud/pubsub/v1/data.proto) seen in prod:

```json
{
  message: {
    data: 'SGVsbG8gdGhlcmU=', 
    messageId: '1408577928008405', 
    publishTime: '2020-08-06T22:31:14.536Z'
  },
  subscription: 'projects/MY-PROJECT/subscriptions/MY-SUB'
} 
```

grant:
Here is an example of a CloudAuditLog seen in prod:
Headers:

```
"host": "my-run-url-uc.a.run.app", 
"content-type": "application/json; charset=utf-8", 
"content-length": "1210", 
"accept": "application/json", 
"from": "noreply@google.com", 
"user-agent": "APIs-Google; (+https://developers.google.com/webmasters/APIs-Google.html)", 
"x-cloud-trace-context": "c0aaeref90d944452f56027d27af0e97/1411812217150857728", 
"x-forwarded-for": "64.233.172.30", 
"x-forwarded-proto": "https",
"forwarded": "for='64.233.172.30';proto=https", 
"accept-encoding": "gzip,deflate,br", 
"ce-source": "//storage.googleapis.com/projects/my-project", 
"ce-time": "2020-08-08T00:11:44.895529672Z", 
"ce-specversion": "1.0", 
"ce-type": "com.google.cloud.auditlog.event", 
"ce-dataschema": "type.googleapis.com/google.logging.v2.LogEntry", 
"ce-id": 
"projects/my-project/logs/cloudaudit.googleapis.com%2Factivity-h6wqfbe7zeke1596845504895522", 
"ce-subject": 
"storage.googleapis.com/projects/_/buckets/my-bucket" } 
```

Payload:

```js
{
  protoPayload: {
    '@type': 'type.googleapis.com/google.cloud.audit.AuditLog',
    status: {},
    authenticationInfo: {
      principalEmail: 'my-email'
    },
    requestMetadata: {
      callerIp: '67.133.97.105;,
      callerSuppliedUserAgent:
[...]

PR #44, opened 2020-08-11
docs: add example directory with cURL requests

grant:
Adds sample HTTP requests with JSON data as seen from Cloud Run.

Folks can simply copy the JSON data if they don't want the full request/headers.

Fixes https://github.com/googleapis/google-cloudevents/issues/39


README.md
```diff
@@ -11,6 +11,9 @@ The "source of truth" for the event data format is in the form of
 See [proto/README.md](proto/README.md) for more
 information about the messages used to describe CloudEvents.
 
+See [examples/README.md](proto/README.md) for examples of
```
jskeet, line 14:
```suggestion
See [examples/README.md](examples/README.md) for examples of
```

examples/README.md
```diff
@@ -0,0 +1,8 @@
+# Example CloudEvents Data in JSON format with HTTP Headers
```
jskeet, line 1: While having cURL scripts is useful, I think it would also be useful to have JSON representations of the data part as plain JSON files too. That would be a lot easier to consume in tests etc. Over time we should probably work out a way of having multiple formats and then a system to generate all those formats from a single request. Obviously that can wait though.

examples/google.events.cloud.audit.v1.md
```diff
@@ -0,0 +1,34 @@
+## google.events.cloud.audit.v1
```
jskeet, line 1: Any reason for having this as a Markdown file rather than just a shell script? Shell scripts still render nicely in GitHub, and that way anyone who has cloned the repo could just call the shell script directly rather than copying it into a new file.

examples/google.events.cloud.pubsub.v1.md
```diff
@@ -0,0 +1,64 @@
+## google.events.cloud.pubsub.v1
```
jskeet, line 1: This doesn't look like a Pub/Sub example - this is a CAL event.

jskeet: (Whereas the audit one *is* a Pub/Sub event - basically the files are just the wrong way round.)

examples/google.events.cloud.audit.v1.md
```diff
@@ -0,0 +1,34 @@
...
+-H "ce-id: 1413058901901494" \
+-H "ce-source: //pubsub.googleapis.com/projects/MY-PROJECT/topics/MY-TOPIC" \
+-H "ce-specversion: 1.0" \
+-H "ce-type: com.google.cloud.pubsub.topic.publish" \
+-H "ce-time: 2020-12-08T20:03:19.162Z" \
+--data-binary @- << EOF
+{
+  "data": {
+    "message": {
+      "data": "SGVsbG8gdGhlcmU=",
```
jskeet, line 26: I think it might be useful to have an example that has attributes as well as data.

README.md
grant: I believe it would be a lot easier to maintain and update these samples if the contents were next to the protos, so I've done that.

examples/google.events.cloud.audit.v1.md
grant: I'd like to have a doc to be able to put documentation about the event.

This product really really needs docs and explanation. We can figure out a better automated testing strategy later, but I think at minimum, we should create a great copy-paste experience, which this provides.

examples/google.events.cloud.pubsub.v1.md
grant: Fixed this.

examples/README.md
grant: Added plain JSON.

jskeet:
Now on vacation - will get back to this on Monday.

grant:
No problem. Cya Monday!

examples/google.events.cloud.audit.v1.md
grant: Sounds fine, but I'd rather generate a Pub/Sub message with this at a later point. This is from a real Pub/Sub message and is valid.

Eventually we should have multiple examples like seen with https://github.com/GoogleCloudPlatform/functions-framework-conformance/tree/master/events/generate/data.

grant:
Also adding @grayside for an additional perspective.

proto/google/events/cloud/audit/v1/README.md
```diff
@@ -0,0 +1,25 @@
+## google.events.cloud.audit.v1
```
grayside, line 1: Having these examples is great! What are the various ways we hope they get used? For example, might we want to generate reference documentation or link this from IDEs?

As we think about shell/protocol samples, I try to remember this is the primary reference for developers in every language for which we don't provide a library.

grant:
Can change directory and merge after #52 is merged.

proto/google/events/cloud/audit/v1/README.md
grant: I'm going to dig up a couple ideas of having a more common reference for curl samples. For now, having them somewhere is better than what we have today.

Talked to Lawrence about this today. Open to ideas!

grant:
I'm going to merge this for now. We have other markdown files in the `proto` directory already. There's a lot of benefit for keeping files not too distant from each other.

Separating the curl command from the docs is something we can likely do, but it's also good to see an overview in the README.

Example:
https://github.com/googleapis/google-cloudevents/blob/master/proto/google/events/cloud/pubsub/v1/README.md

PR #46, opened 2020-08-12
Firebase data schema fixes

jskeet:
The first commit should be absolutely fine. The second I'm slightly less happy about, in that I'd like to see all those fields come up. (If you have done any Firebase work and can help on that front, that would be great. Otherwise, we'll just talk with the Firebase folks.)


jskeet:
Ping?

PR #47, opened 2020-08-12
Prototype of changing the CAL handling

jskeet:
Two significant changes here:

- There's an extra level of wrapping, as the CloudEvent data is a `LogEntry`, not an `AuditLog`
- Introduced an extra somewhat-synthetic field into `AuditLog` with a JSON key of "@type"

This is incomplete - it's just a prototype for test purposes (as there's more data in LogEntry) but I've tried generating the code with C#, and yes, we can retrieve the value from the `@type` field in the JSON.
If `AuditLog` is now packed into an `Any` and converted to JSON (with an appropriate type registry), we now get two `@type` fields in the JSON output, which is basically broken... but we *could* decide that's an acceptable edge case.

### DIFF
--- a/proto/google/events/cloud/audit/v1/data.proto
+++ b/proto/google/events/cloud/audit/v1/data.proto
@@ -17,16 +17,31 @@ syntax = "proto3";
 package google.events.cloud.audit.v1;
 
 import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
 import "google/rpc/context/attribute_context.proto";
 import "google/rpc/status.proto";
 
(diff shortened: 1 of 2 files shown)


jskeet:
(Note that "original_protobuf_type" is an arbitrary choice of field name here. We'd want something we can use consistently, but that won't clash with other, genuine fields.)

jskeet:
I'll leave this PR as it is for further discussion around @type, but I'm creating a new one that *just* has the wrapping part.

jskeet:
(#48 supercedes this)

PR #48, opened 2020-08-17
Wrap Cloud Audit Logs in a LogEntry proto

jskeet:
(Replaces the prototype in #47.)

This is ready for review, but we should dig more carefully to work out how to answer the questions in the second commit.

This PR does *not* include anything that will interpret the `@type` field that is emitted as a byproduct of the `Any` field.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,22 @@ This repository contains definitions for the following CloudEvents:
 
 |Package|Event types|Data messages|
 |-|-|-|
+|[google.events.cloud.audit.v1](proto/google/events/cloud/audit/v1)|google.cloud.audit.log.v1.written|LogEntryData|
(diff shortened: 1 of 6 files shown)


jskeet:
> How do we verify that the CloudEvent data shape seen here matches what we actually see in production?

That's orthogonal to this PR. Yes, I'm aware that we need validation of this, but that needs a separate design, and I'd expect a PR for that to include all kinds of aspects that aren't changes to the schema.
We can hold off on this change until we've got the testing infrastructure, but that will leave this "wrong" for longer.

(Not that I'm planning on merging it immediately anyway.)

jskeet:
(Will start on a separate PR for that test framework though.)

yolocs:
Another route (not sure if possible): if events could be marshaled use generated lib, then they will for sure have the schemas defined here.

jskeet:
Merging now after internal confirmation.

Issue #50, opened 2020-08-17
Verify Proto Definitions of JSON Payloads Match JSON Payloads Seen by Cloud Run


grant:
## Expected Behavior

There should be no difference between the payload definitions seen by the protos in this repo (and their JSON representations) with the HTTP body seen by requests sent via Events for Cloud Run. This ensures that we generate valid event type libraries that are compatible with the payloads we see in the service.

The point of this repo is define these types.

To ensure there is no difference in these JSON objects, we should at least have actual JSON payloads seen by the service and ideally have e2e tests verify the definitions match what's seen in production.

I'm not sure how to verify a protobuf definition matches the JSON payload though.

## Actual Behavior

We have no testing and no way to ensure that these type definitions are correct.

CC @jskeet

grant:
Yash is working on validating the protos in this repo match what is produced by Eventarc. (Likely using the Go/protoc).

Jon has done some work here as well.

Given that, I don't think there's API validation needed in this repo. So closing.

PR #52, opened 2020-08-18
Sample data for validation

jskeet:
(Only the last commit is new - the rest is based on #48.)

This is sample data that can be used by code in language-specific repos to validate the schema. Once we've got the relevant bits in place (the data and the code to validate it) we can add it as a continuous integration check.
This is one part of addressing #50.

See https://github.com/googleapis/google-cloudevents-dotnet/pull/30/files for the corresponding code to test this with .NET. (Ideally we should have something similar in every language.)

### DIFF
--- a/testdata/README.md
+++ b/testdata/README.md
@@ -1,0 +1,22 @@
+# Test data
+
+The JSON files within this directory hiearchy represent test data
+for the schema in the repository. The directory structure matches
+that in the [proto](../proto) directory, namely the protobuf package
+hierarchy.
+
+Each JSON file should be named according to the CloudEvent data
+protobuf message it represents, with an optional description suffix
+starting with a `-`. Examples:
+
+- `StorageObjectData.json`
+- `DocumentEventData-simple.json`
+- `DocumentEventData-complex.json`
+
+When the proto schema is updated in a non-trivial way, either a new
+JSON file should be created or an existing file should be modified,
+to validate that the schema change is correct. The test file should
+include whatever fields and messages are being added or modified.
+Ideally, the test file should be created by observing an actual
+CloudEvent, copying the event data, and removing any sensitive
+information. 
--- a/testdata/google/events/cloud/pubsub/v1/MessagePublishedData-binary.json
+++ b/testdata/google/events/cloud/pubsub/v1/MessagePublishedData-binary.json
@@ -1,0 +1,8 @@
+{
+  "subscription": "projects/my-project/subscriptions/my-subscription",
+  "message": {
+    "@type": "type.googleapis.com/google.pubsub.v1.PubsubMessage",
+    "data": "AQIDBA==",
+    "messageId": "message-id"
+  }
+}
--- a/README.md
+++ b/README.md
@@ -62,3 +62,8 @@ running the following commands below a machine with .NET Core SDK
 ./compile.sh
 ./generate-event-registry.sh tmp/protos.pb
 ```
+
+Non-trivial changes to the schema should also include a suitable
+test data file (or a change to an existing test data file) to
+validate the change. See [testdata/README.md](testdata/README.md)
+for details.
--- a/testdata/google/events/cloud/firestore/v1/DocumentEventData-complex.json
+++ b/testdata/google/events/cloud/firestore/v1/DocumentEventData-complex.json
@@ -1,0 +1,71 @@
+{
+  "oldValue": {},
+  "updateMask": {},
+  "value": {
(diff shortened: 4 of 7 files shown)


testdata/google/events/cloud/pubsub/v1/MessagePublishedData-binary.json
```diff
@@ -0,0 +1,8 @@
+{
+   "subscription": "projects/my-project/subscriptions/my-subscription",
```
grant, line 2: Indentation is inconsistent. Should be 2 spaces in all JSON files.

testdata/google/events/cloud/audit/v1/LogEntryData-issue43.json
```diff
@@ -0,0 +1,35 @@
+{
```
grant, line 1: For the file name, I think issue 43 will be long gone from most users.

We should just have something like `LogEntryData-storageBbucketsUpdate` to provide a searchable name.

jskeet: Happy to rename all of these, definitely.

testdata/google/events/cloud/pubsub/v1/MessagePublishedData-binary.json
jskeet: Will do (although probably not this minute).

jskeet:
> Given that this repo just contains definitions for types, and is not actually used in deployed projects, I would expect JSON, Proto, Avro, etc representations of a specific type to be in the same folder. It would be nice to just pull *.json alongside the .proto definition. You commented somewhere on this topic, but I can't find the exact comment.

The comment was on https://github.com/googleapis/google-cloudevents/pull/44, and I still maintain that there's a significant difference between the proto files and the JSON files here - the proto files *effectively* ship as code. We need to be ultra careful with them in terms of breaking changes, for example. The JSON files are "just" sample data, like tests - we can change them as much as we like. I'm very keen on maintaining that separation.

jskeet:
Cool. Will rebase, rename etc - probably tomorrow morning UK time.

jskeet:
Looks like the CAL sample is pretty incomplete, in that it only has data for protoPayload - will find a fuller example.

jskeet:
Note: we're working out exactly the order to do things in.

testdata/README.md
```diff
@@ -0,0 +1,21 @@
+# Test data
```
grant, line 1: How do we generate this test data? We need some rough instructions or information about this test data so that others can reproduce this library if one person is away.

jskeet: That's the next step, and possibly an internal one. I would really like to make progress rather than try to boil the ocean.

testdata/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+The JSON files within this directory hiearchy represent test data
+for the schema in the repository. The directory structure matches
+that in the [proto](../proto) directory, namely the protobuf package
+hierarchy.
+
+Each JSON file should be named according to the data message it
```
grant, line 8: At first I didn't realize the naming aligns to the protobuf message / "Data messages" field as seen in the README.

Can we say something like the "according to the corresponding protobuf's `message`"

testdata/README.md
```diff
@@ -0,0 +1,21 @@
...
+represents, with an optional description suffix. Examples:
```
grant, line 9: optional description suffix starting with a `-`. (Rather than a + or something else).

testdata/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+- `StorageObjectData.json`
+- `DocumentEventData-simple.json`
+- `DocumentEventData-complex.json`
+
+When the proto schema is updated in a non-trivial way, either a new
+file should be created or an existing file should be modified, to
```
grant, line 16: A new _json_ file?

grant:
For progress, we can merge these PRs.

testdata/README.md
jskeet: I think that's reasonably implicit here, but it does no harm. Added.

testdata/README.md
jskeet: Done

testdata/README.md
jskeet: Changed to "CloudEvent data protobuf message" which I think its unambiguous.

jskeet:
Merging this without the CAL sample, which is moved to #48.

PR #53, opened 2020-08-19
Migrate to GitHub actions and add schema validation

jskeet:


jskeet:
See https://github.com/jskeet/google-cloudevents/runs/1002434751 (if you have access...) for the GitHub Actions log.
(I assume that in future PRs, it would show up in the checks here too.)

PR #57, opened 2020-08-26
fix: Move the CloudBuild events under google.events.cloud.cloudbuild

jskeet:
This makes the proto structure match the event type.
(The duplication of "cloud" in the name has been discussed
separately.)

cc @yolocs. @capri-xiyue


jskeet:
The point of this PR isn't to be consistent with the DNS name for the API. It's to be consistent with the event type, which Scheduler already is.

bendory:
LGTM for `cloudbuild`

yolocs:
/lgtm

This was definitely a miss in the first place :) Thanks for taking care of this.

PR #59, opened 2020-08-26
Add JSON schemas for Pub/Sub and Audit Log events

michaelawyu:
Note: The two JSON schemas added are for experimentation purposes only; they are handwritten based on the `.proto` definitions and used only for proof-of-concept typing library generation.


proto/google/events/cloud/audit/v1/events.json
```diff
@@ -0,0 +1,189 @@
+{
+    "$schema": "http://json-schema.org/schema",
+    "title": "google.events.cloud.audit.v1",
+    "goPackage": "auditv1",
```
grant, line 4: I don't think we should have language-specific  keys in our schema, right?

proto/google/events/cloud/pubsub/v1/events.json
```diff
@@ -0,0 +1,25 @@
+{
```
grant, line 1: How did we create this?

I'd imagine we'd use quicktype to generate the basic schemas, then tweak them a bit, something like this:

```json
{
    "$schema": "http://json-schema.org/draft-06/schema#",
    "$ref": "#/definitions/PubSubData",
    "definitions": {
        "PubSubData": {
            "type": "object",
            "additionalProperties": false,
            "properties": {
                "message": {
                    "$ref": "#/definitions/Message"
                },
                "subscription": {
                    "type": "string"
                }
            },
            "required": [
                "message",
                "subscription"
            ],
            "title": "PubSubData"
        },
        "Message": {
            "type": "object",
            "additionalProperties": false,
            "properties": {
                "data": {
                    "type": "string"
                },
                "messageId": {
                    "type": "string"
                },
                "publishTime": {
                    "type": "string",
                    "format": "date-time"
                },
                "attributes": {
                    "$ref": "#/definitions/Attributes"
                }
            },
            "required": [
                "data",
                "messageId",
                "publishTime"
            ],
            "title": "Message"
        },
        "Attributes": {
            "type": "object",
            "additionalProperties": false,
            "required": [],
            "title": "Attributes"
        }
    }
}
```

proto/google/events/cloud/audit/v1/events.json
michaelawyu: It is true that `goPackage` is not a field listed in the JSON schema spec; unfortunately we need it to generate the Go package (in the same way as Protocol Buffers Generator requires the `go_package` option). Alternatively we could infer this value from the `title` field for languages like Go, but it will not be very elegant (have to use hard coded prefixes).

proto/google/events/cloud/pubsub/v1/events.json
michaelawyu: Yeah, it would work. We have to find a way to check the generated schema against the `.proto` files though :(

proto/google/events/cloud/audit/v1/events.json
grant: No issues with this then :)

PR #60, opened 2020-08-29
feat: add quicktype tool

grant:
# Feat: Add quicktype tool

- Updates JSONSchema
  - `$id`: unique identifier for the schema
  - `title`: the human readable title
  - `name`: the name of this data type
- Moves registry tool in separate directory in `tools/`
- Inits TypeScript project, `tools/quicktype-wrapper/`
  - Uses [`google/gts`](https://github.com/google/gts) for basic automatic styling
  - Crawls a directory, and produces equivalent language files for every `schema.json` file.
  - Adds basic README for the tool.
- Tool can be run as a Node script or installed as a CLI.

## Example Output

```
***********************
** Quicktype Wrapper **
***********************
***********************
* Environment Variables:
- IN=/Users/timmerman/Documents/github/googleapis/google-cloudevents/proto/
- OUT=/Users/timmerman/Documents/out
- L=TYPESCRIPT
***********************

== START ==
- Finding all JSON Schemas (schema.json)...
Found 2 schema(s).
- LogEntryData                            : google/events/cloud/audit/v1/LogEntryData.ts
- MessagePublishedData                    : google/events/cloud/pubsub/v1/MessagePublishedData.ts
== END ==
```


jskeet:
Will look at this on Tuesday (Monday is a public holiday in the UK).

grant:
Requesting a review from Adam too.

tools/quicktype-wrapper/README.md
```diff
@@ -2,16 +2,37 @@
...
+This tool can be installed as a global CLI or run as a Node script. Requires Node 8+.
+
+### CLI (Recommended)
+
+Install the CLI globally:
 
 ```sh
-npm run build
-npm start
+sudo npm run cli
```
grayside, line 14: General practice is to never use sudo to install an npm packge. Why is this necessary?

tools/quicktype-wrapper/README.md
```diff
@@ -2,16 +2,37 @@
...
+
+```sh
+npm run compile
+npm run start
+```
+
+## Run
+
+const IN = process.env.IN || `${HOMEDIR}/Documents/github/googleapis/google-cloudevents/proto/`;
+const OUT = process.env.OUT || `${HOMEDIR}/Documents/out`;
```
grayside, line 31: Working in Documents seems odd? I assume github/googleapis/google-cloudevents/proto is coming from a separately git cloned repository?

tools/quicktype-wrapper/README.md
grant: This is necessary for some folks because you're installing a bin `qt` globally. Otherwise you'll see:

```
Error: EACCES: permission denied, access '/usr/local/lib/node_modules'
# or
Error: EACCES: permission denied, unlink '/usr/local/bin/qt'
```

tools/quicktype-wrapper/README.md
grant: These are some default values that if you don't specify one.
(I recommend we always specify one)

It's nice being able to just execute `qt`.

We could consider not having defaults or just writing an error, but working in `Documents` rather than cluttering ~ is better imo.

jskeet: Is it normal in Node to use environment variables like this, rather than providing them as command-line arguments? In other environments I'd *definitely* be using command line arguments, and I wouldn't assume any defaults - it feels to me like the chances of the defaults being right for more than a single person are pretty small.

tools/quicktype-wrapper/README.md
```diff
@@ -0,0 +1,50 @@
...
+To continually watch the script insead of building after every change,
+run this command in a separate terminal:
+
+```sh
+npm run watch
+```
+
+## Lint/Fix
+
+To fix style/lint mistakes, run:
```
jskeet, line 46: What is this to fix mistakes in - the JSON schema files?

jskeet (commented):
Just a couple of general comments. I'm not qualified to review the Typescript, but I'm happy enough with the general idea.

tools/quicktype-wrapper/README.md
grant: No, this README is in the `tools/quicktype-wrapper/` folder/project. The JSON schemas are not in this folder (in this repo, but really tangential).

jskeet: So what *would* it fix? If this is just generating code, wouldn't any style/lint issues suggest a problem with the generator or the schema?

tools/quicktype-wrapper/README.md
grant: The goal of this tool is not to build a CLI to generate, it's to use the `npm run gen` script.

I think I can just remove these defaults.

tools/quicktype-wrapper/README.md
grant: Fixing the schema is out of scope of this PR/folder. Any tooling to do that would not live in the `tools/quicktype-wrapper` tool.

Why do you want this feature here?

jskeet: I don't particularly - I'm trying to find out what this *does* fix mistakes in; that's what I've been asking.

grant: Added a more detailed comment.

tools/quicktype-wrapper/README.md
grant: Removed defaults.

grant:
Ready for another review. Hoping to merge soon.

tools/quicktype-wrapper/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Here's an example:
+
+```sh
+IN=~/Documents/github/googleapis/google-cloudevents/proto
+OUT=~/Documents/out
+L=typescript
+```
+
+Then run `qt`, which will pick up these environment variables.
```
jskeet, line 40: Again, I find it odd to use environment variables rather than command line arguments, but maybe that's just a Node convention.

jskeet (approved):
That's definitely clearer, thanks. I suspect I won't need to use this, at least for a while, so it makes sense to aim it at usability for you for now - and I'm definitely in favour of merging and iterating.

tools/quicktype-wrapper/README.md
grant: Removed it fyi.

tools/quicktype-wrapper/README.md
grant: Will create a GitHub issue.

grant:
Filed https://github.com/googleapis/google-cloudevents/issues/61 per above question.

Issue #61, opened 2020-09-01
Support CLI flags in addition to Env Vars for Quicktype Tool


grant:
## Expected Behavior

The quicktype tool should support CLI flags.

```
qt --in /foo/bar --out /baz --lang php;
```

## Actual Behavior

Currently you must provide env vars.

```
OUT=...
IN=...
LANG=...
qt;
```

@grayside @jskeet

grant:
See PR.

PR #63, opened 2020-09-02
feat: enable cli flags to quicktype wrapper

grant:
Adds CLI flags to quicktype tool

Tested locally.

```
qt \
--in /Users/timmerman/Documents/github/googleapis/google-cloudevents/proto/ \
--out /Users/timmerman/Documents/github/googleapis/google-cloudevents-nodejs/src
***********************
** Quicktype Wrapper **
***********************
***********************
* Config:
- IN=/Users/timmerman/Documents/github/googleapis/google-cloudevents/proto/
- OUT=/Users/timmerman/Documents/github/googleapis/google-cloudevents-nodejs/src
- L=TYPESCRIPT
***********************

== START ==
- Finding all JSON Schemas (schema.json)...
Found 2 schema(s).
- LogEntryData                            : google/events/cloud/audit/v1/LogEntryData.ts
- MessagePublishedData                    : google/events/cloud/pubsub/v1/MessagePublishedData.ts
== END ==
```

Fixes: https://github.com/googleapis/google-cloudevents/issues/61

### DIFF
--- a/tools/quicktype-wrapper/src/index.ts
+++ b/tools/quicktype-wrapper/src/index.ts
@@ -2,18 +2,21 @@
 
 import { readFileSync, writeFileSync } from "fs";
 import {jsonschema2language, LANGUAGE, LANGUAGE_EXT, TARGET_LANGUAGE} from './quickstype';
+const {argv} = require('yargs')
 const mkdirp = require('mkdirp');
 const HOMEDIR = process.env[(process.platform == 'win32') ? 'USERPROFILE' : 'HOME'];
 
 /**
  * A simple tool that generates code using quicktype.
+ * 
+ * Configuration (via environment variables or command-line flags):
  * @param {string} IN The directory for JSON Schema input. Must have trailing /.
  * @param {string} OUT The directory for generated output. Must have trailing /.
  * @param {string} L The target language
  */
-const IN = process.env.IN;
-const OUT = process.env.OUT;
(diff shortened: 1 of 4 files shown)


tools/quicktype-wrapper/src/index.ts
```diff
@@ -2,26 +2,29 @@
...
+ * 
+ * Configuration (via environment variables or command-line flags):
  * @param {string} IN The directory for JSON Schema input. Must have trailing /.
  * @param {string} OUT The directory for generated output. Must have trailing /.
  * @param {string} L The target language
  */
-const IN = process.env.IN;
-const OUT = process.env.OUT;
-const L = (process.env.L || LANGUAGE.TYPESCRIPT).toUpperCase() as TARGET_LANGUAGE;
+const IN = process.env.IN || argv.in;
```
jskeet, line 17: My guess is that this means the environment variable takes precedence over the command line argument - is that the case? If so, I'd suggest reversing it - treat the environment variable as a default, allowing the command line argument to override it.

grant: Yeah, should switch this.

jskeet (approved):
I can't really comment on the Typescript quality, but it looks like it's doing the right thing. (I would probably have used "lang" or similar for the command line argument name and GENERATOR_LANG for the environment variable, but that's personal preference.)

PR #68, opened 2020-09-10
feat: add descriptions to main jsonschema fields

grant:
Adds descriptions for many fields in the JSON Schemas for better IDE results. Descriptions pulled from protos/docs.

Not comprehensive.

### DIFF
--- a/proto/google/events/cloud/audit/v1/schema.json
+++ b/proto/google/events/cloud/audit/v1/schema.json
@@ -22,17 +22,30 @@
                 },
                 "ResourceLocation": {
                     "type": "object",
+                    "$comment": "https://cloud.google.com/service-infrastructure/docs/service-control/reference/rpc/google.cloud.audit?hl=en#resourcelocation",
                     "properties": {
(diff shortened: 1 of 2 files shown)


jskeet (approved):
I'm fine with these, but I don't know whether we'll be able to automate the same level of detail - particularly around examples - without pretty significant work. I wonder how wise it is to "over-deliver" on some items now, when we won't be able to scale that.

Having said that, for the purpose of *seeing the effect* of such comments (so we can work out how much effort it's worth putting into the schema generator when we get to that) I'm fine with this. I just think we probably need to focus on automation next instead of trying to make this comprehensive in more PRs.

grant:
Thanks.
I'm assuming **GCF types** can at least have these fields.

I would also love automation, but I think it's worth while seeing the effects of these comments in the interim. I think we'd need an advanced proto parser + some manual additions if we were to automate this.

I am not concerned about over-delivering comments/examples.

Going to merge this and test out the result.

PR #80, opened 2020-10-01
feat: add version numbers to jsonschema and update qt tool

grant:
Updates the JSON Schema wrapper tool (`qt`) to use separate dedicated jsonschema directory (see https://github.com/googleapis/google-cloudevents/pull/77). The old generator doesn't work with the new directory.

- Simplifies qt logic a bit and adds more comments. Diff is pretty small.
- Makes a couple small fixes to the schemas
  - Remember to add the version number in the directory (`v1`)
  - Adds a missing description for Pub/Sub.

Tested with Node. Example for Node:

```
***********************
** Quicktype Wrapper **
***********************
***********************
* Config:
- IN=/Users/timmerman/Documents/github/googleapis/google-cloudevents/jsonschema
- OUT=/Users/timmerman/Documents/github/googleapis/google-cloudevents-nodejs
- L=TYPESCRIPT
***********************

== START ==
- Finding all JSON Schemas (*.json)...
Found 2 schema(s).
- LogEntryData                            : cloud/audit/v1/LogEntryData.ts
- MessagePublishedData                    : cloud/pubsub/v1/MessagePublishedData.ts
== END ==
```

Mostly looking for 1 LGTM and get this in, but other comments are welcome.


tools/quicktype-wrapper/src/index.ts
```diff
@@ -91,27 +90,29 @@ if (!module.parent) {
     // Loop through every path
     const pathPromises = absPaths.map(async (f: string, i: number) => {
       const file = await readFileSync(f) + '';
-      const relPath = relPaths[i];
-      const typeName = JSON.parse(file).name;
+      const relPath = relPaths[i]; // e.g. /google/events/cloud/pubsub/MessagePublishedData.json
```
jskeet, line 93: I'm confused by this being called relPath (which I assume to mean "relative") but with an example starting with "/" which sounds absolute. Is the example correct, or would it actually start with google?

jskeet (approved):
Basically fine as far as I can see, although I'm really not competent at Typescript

tools/quicktype-wrapper/src/index.ts
grant: Good point. Yeah it's relative, it's all the stuff after "googleapis/google-cloudevents/jsonschema". Naming is hard. Probably should rename to something else or change the logic. Path separators are a bit tedious to keep track of with all these generators.

tools/quicktype-wrapper/src/index.ts
```diff
@@ -19,16 +19,16 @@ const OUT = argv.out || process.env.OUT;
...
 /**
  * Gets a list of all JSON Schema paths (absolute paths)
  * @param directory The path to the directory with schemas.
  */
 async function getJSONSchemasPaths(directory: string) {
-  console.log(`- Finding all JSON Schemas (${SCHEMA_FILE})...`);
+  console.log(`- Finding all JSON Schemas (${SCHEMA_PATTERN})...`);
   const recursive = require("recursive-readdir");
-  const paths: string[] = await recursive(directory, [`!${SCHEMA_FILE}`])
+  const paths: string[] = await recursive(directory, [`!${SCHEMA_PATTERN}`]);
```
grayside, line 31: nit: This syntax was sufficiently obtuse I researched the library. Maybe a comment on excluding files without this pattern?

tools/quicktype-wrapper/src/index.ts
grayside: Given the example, this might be more readable with a topical name.

E.g., pathToSchema

grayside (approved):

tools/quicktype-wrapper/src/index.ts
grant: Good point. I extracted this pattern and added a comment.

grant:
Addressed @grayside's comments with changes in the last non-merge commit.

PR #82, opened 2020-10-02
fix: change cal jsonschema to use camelCase

grant:
Modifies the JSON Schema field keys to use camelCase as provided by the Eventarc API, but not the representation used in protos.

See:
- go/proto3-json-spec
- https://developers.google.com/protocol-buffers/docs/proto3#json_options

### DIFF
--- a/jsonschema/google/events/cloud/audit/v1/LogEntryData.json
+++ b/jsonschema/google/events/cloud/audit/v1/LogEntryData.json
@@ -12,11 +12,11 @@
       "description": "The monitored resource that produced this log entry.",
       "properties": {
         "type": {
-          "description": "Required. The monitored resource type. For example, the type of a Compute Engine VM instance is `gce_instance`.",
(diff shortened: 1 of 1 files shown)


grayside:
Per comment in https://github.com/googleapis/google-cloudevents-nodejs/pull/54, any stable libraries regenerated from this schema will have compatibility breaks. Not sure if this should be noted in the repo.

grant:
> Per comment in [googleapis/google-cloudevents-nodejs#54](https://github.com/googleapis/google-cloudevents-nodejs/pull/54), any stable libraries regenerated from this schema will have compatibility breaks. Not sure if this should be noted in the repo.

Well the JSON Schema for CAL was incorrect. The JSON Schemas are not the source of truth.

This change fixes the JSON Schemas to a correct state, as the JSON payload does not use underscore_case. Google APIs data payloads definitely don't use that casing.

There are no stable libraries that use this schema. I do not think there is anything extra to note.

michaelawyu (approved):

grayside (approved):

PR #89, opened 2020-10-12
feat: qt enable headers flag

grant:
Per request, adds option to add license headers to generated files.

Fixes: https://github.com/googleapis/google-cloudevents/issues/88

### DIFF
--- a/tools/quicktype-wrapper/README.md
+++ b/tools/quicktype-wrapper/README.md
@@ -29,11 +29,16 @@ npm run start
 
 To configure the script, use _environment variables_ or _command-line flags_:
 
+- (REQUIRED) `IN`: Input directory path
+- (REQUIRED) `OUT`: Output directory path
+- (NOT REQUIRED) `NO_LICENSE`: Adds a license header to source files
+
 ### Environment Variables
 
 ```sh
 IN=~/Documents/github/googleapis/google-cloudevents/proto
 OUT=~/Documents/out
+NO_LICENSE=true # optional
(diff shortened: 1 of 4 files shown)


tools/quicktype-wrapper/README.md
```diff
@@ -29,11 +29,16 @@ npm run start
 
 To configure the script, use _environment variables_ or _command-line flags_:
 
+- (REQUIRED) `IN`: Input directory path
+- (REQUIRED) `OUT`: Output directory path
+- (NOT REQUIRED) `NO_LICENSE`: Adds a license header to source files
```
grayside, line 34: Why need it to be optional?

grayside, line 34:
```suggestion
- (NOT REQUIRED) `NO_LICENSE`: Set to `true` to skip adding license headers to source files.
```

grayside (approved):
LGTM. Asked a question and made a nit suggestion but neither should block a clean change.

tools/quicktype-wrapper/README.md
grant: Was thinking that some languages might want to add their own license headers or add them in a different way. Opt-out sounds fine for that.

grant: Fix applied, thanks.

averikitsch (approved):

cum1997:
16a3644614b00cbf30a2545b9998e34f0662ff5f

PR #90, opened 2020-10-13
refactor: move json examples to separate folder

grant:
- Renames `testdata` to `json` to have one consistent place for json samples.
- Moves the example JSON files in the `proto` directory to this folder.

Now we have:
- `json`
- `jsonschema`
- `proto`

representations for our CloudEvents, with the same folder structure, completely separate.

---

It seems like this repo has a dependency on the .NET client? Well, perhaps we can have that repo checkout a previous version of this repo, and once we merge this, change that test back to use master.

---

I hope to embed some of these JSON samples on CloudSite.


json/README.md
```diff
@@ -1,4 +1,4 @@
-# Test data
+# JSON Example Data
```
jskeet, line 1: There's a difference between example data and test data. There's definitely overlap, but examples are less likely to exhaustively use corner cases, for example.
It's not clear to me which kind of data this directory is meant to have, given that the heading doesn't match the description in the first sentence.

If we want it to cover both "examples we want humans to read to understand the event structure" and "test data that we want machines to read for validation" then we'll need a convention for distinguishing between them... or we could split them into two directories. But I definitely want it to be clear for any given file whether this is something that a user is likely to learn from, or whether it's primarily there for testing purposes.

jskeet: I'm very much opposed to that approach. If a user comes to this directory and wants to find a sample file which is a reasonable example of what they might expect in (say) a CAL, which one do they look at, if in 6 months we've got 100 CAL samples that test all kinds of corner cases (which seems reasonable to me). When *writing* a JSON file, I want to pay attention to whether the aim is teaching humans or testing code - I'm unlikely to get the best outcome for both of them.

So I'd say either we only want one kind of file, or it's best to split them up into different directories.

grant: I don't understand these thoughts – perhaps we can sync more next meeting.

> I'm very much opposed to that approach. If a user comes to this directory and wants to find a sample file which is a reasonable example of what they might expect in (say) a CAL, which one do they look at, if in 6 months we've got 100 CAL samples that test all kinds of corner cases (which seems reasonable to me).

Any 1 of the 100 samples are valid CAL JSON representations. I don't think folks are going to randomly stumble upon a folder with json via GitHub and be confused in 6 months.

> When _writing_ a JSON file, I want to pay attention to whether the aim is teaching humans or testing code - I'm unlikely to get the best outcome for both of them.
> 
> So I'd say either we only want one kind of file, or it's best to split them up into different directories.

How about we have machine-readable files in different formats. Proto is one, JSON Schema is one, JSON is another. To match that, we can simply rename the folder from `testdata` to `json` to match the fact that the contents is pure json.

I don't want to maintain two folders/sets of machine-readable JSON payloads that overlap. And I'd like to have the folder name to be consistent.

jskeet: > I don't want to maintain two folders/sets of machine-readable JSON payloads that overlap. 

Even if they're being created for radically different reasons, with very different actual payloads? While any useful example should also be a valid machine-readable JSON payload, the reverse is not true. A file deliberately trying to provoke corner cases is less likely to be useful for learning purposes IMO.

As you say, let's talk more at the next meeting.

Issue #74, opened 2020-09-25
Validate JSON Schemas against Protos


grant:
## Expected Behavior

Protos are the source of truth, but JSON schemas are used by many clients. We should validate that the JSON Schemas conform to the correct types that the protobufs (data protobufs).

For each JSON schema file, verify that the corresponding protobuf file represents the same data structure.

## Actual Behavior

We don't have any validation yet.

grant:
Yash is working on this internally.

grant:
Related, @grayside has done some great work with JSON schema generation:
https://gist.github.com/grayside/4d2e7cf67234248b6c70cd7fc6c1655a#run-the-generator-for-audit-logs

grant:
Or ideally generate the jsonschemas or merge the generated schemas with current schemas:

https://github.com/chrusty/protoc-gen-jsonschema

grant:
So, it's likely we should just gen the jsonschemas and then validate schemas against JSON payload examples. There are many tools to validate that, like:

https://www.npmjs.com/package/jsonschema

grant:
We now gen from protos, so no validation needed here.

PR #93, opened 2020-10-26
Init proto2jsonschema

grant:
## Init proto2jsonschema

- Adds `proto2jsonschema` tool
  - Loops through `data.proto` files
  - Uses jsonschema proto plugin: https://github.com/chrusty/protoc-gen-jsonschema
  - Runs postgen step
    - Removes duplicate snake_case fields (there's an [upstream GitHub issue](https://github.com/chrusty/protoc-gen-jsonschema/issues/38) to natively remove these duplicate fields).
    - Adds `name` and `$id` properties in postgen step (on top of file)
- Includes `README.md` for this tool
- Runs the generator

Issue: https://github.com/googleapis/google-cloudevents/issues/74
R: @michaelawyu @jskeet 
CC: @grayside


.gitignore
```diff
@@ -4,5 +4,7 @@ tmp/
 .vs/
 .idea/
 
+proto/googleapis
```
jskeet, line 7: I wouldn't expect this to be necessary - git knows what a submodule is, and shouldn't be trying to add it as a normal file.

.gitignore
```diff
@@ -4,5 +4,7 @@ tmp/
...
+
```
jskeet, line 8: (Adding here as I can't add a comment on the submodule add itself...)

It's not clear to me yet (and hopefully it'll *become* clear) why we've got the submodule at all. The intention of `third_party/googleapis` was to capture enough of `googleapis` for the events, without inadvertently allowing undesirable dependencies.

If we *do* want it, I think it would be better in `third_party` or some other directory separated from `proto` - having `proto/google` and `proto/googleapis/google` as the logically-equivalent proto paths is odd.

testdata/README.md
```diff
@@ -1,22 +0,0 @@
-# Test data
```
jskeet, line 1: Why has this directory been removed? (I don't see why "generating the JSON schema" and "removing test data" would be in the same pull request in general, and I question whether it's a good idea. I use that test data in google-cloudevents-dotnet.)

tools/proto2jsonschema/README.md
```diff
@@ -0,0 +1,30 @@
...
+From the root directory of this repo, run this script:
+
+```sh
+./tools/proto2jsonschema/gen.sh
+```
+
+This script:
+- Installs protobuf tools: `protoc`
+- Installs protobuf extension for JSON Schemas: `chrusty/protoc-gen-jsonschema`
+- Clones `googlapis/googleapis`
```
jskeet, line 23: Why would we clone this when a) you've introduced a submodule for it and b) all the required protos are expected to be in `third_party/googleapis`?

tools/proto2jsonschema/gen.sh
```diff
@@ -0,0 +1,106 @@
...
+go get -v github.com/chrusty/protoc-gen-jsonschema/cmd/protoc-gen-jsonschema
+go install github.com/chrusty/protoc-gen-jsonschema/cmd/protoc-gen-jsonschema
+
+echo "- Converting protos to JSON Schemas"
+chmod +x $PROTOC
+
+# Clone googleapis dependency
+cd proto
+if [ ! -d "googleapis" ] ; then
+  git clone https://github.com/googleapis/googleapis
```
jskeet, line 53: This is cloning in the same place you've got a submodule - that would leave the repository in an inconsistent state.
Again, I don't think we need the submodule at all, but if we *do* want it, this script should expect it to be there as a submodule. If it's not there, error out with a message about using `git submodule init` etc.

jskeet (commented):
I haven't reviewed the actual JSON schema output, but I'd really like to avoid the submodule/clone of googleapis. It shouldn't be necessary, and it would allow undesirable dependencies. I'd also like to keep the testdata directory.

testdata/README.md
grant: Sorry, this shouldn't have been removed. It would be out of scope anyways if removed.
Reverted.

.gitignore
grant: There should be no cloning of googleapis, this was never used as part of the `protoc` generation process and was a mistake adding it. This instruction is outdated.

.gitignore
grant: There should be no cloning of googleapis, this was never used as part of the `protoc` generation process and was a mistake adding it. This instruction is outdated.

tools/proto2jsonschema/gen.sh
grant: There should be no cloning of googleapis, this was never used as part of the `protoc` generation process and was a mistake adding it. This instruction is outdated.

tools/proto2jsonschema/README.md
grant: There should be no cloning of googleapis, this was never used as part of the `protoc` generation process and was a mistake adding it. This instruction is outdated.

grant:
Thanks for reviewing @jskeet. I reverted the mistake of removing `testdata` and I resolve the comments about cloning googleapis, which was never used.

jskeet:
This still adds `googleapis` as a submodule, which I'd prefer it didn't. Is there a reason you decided to add it rather than using the existing `third_party/googleapis` directory? If we *do* want to change to use a submodule, I'd suggest using the existing location and just replacing the copied files with the submodule.

grant:
> This still adds `googleapis` as a submodule, which I'd prefer it didn't. Is there a reason you decided to add it rather than using the existing `third_party/googleapis` directory? If we _do_ want to change to use a submodule, I'd suggest using the existing location and just replacing the copied files with the submodule.

Pardon, I now see that in the PR and it has been removed. Again, this is not used in the generation script/process.

jskeet (approved):
Changes look fine now, thanks - I think for the sake of the history, it would be worth rebasing locally and force pushing. While you could still split that up into multiple commits, I think it would be fine as just one commit.

tools/proto2jsonschema/gen.sh
```diff
@@ -0,0 +1,99 @@
...
+# Only run in repo root
+currentDir=${PWD##*/}
+
+if [ "$currentDir" != "google-cloudevents" ] ; then
+  echo "Error: Script must be run in google-cloudevents root.";
+  exit 1;
+fi
+
+# Install protoc
+case "$OSTYPE" in
```
grayside, line 14: What if protoc is already installed? Or is this just for CI?

tools/proto2jsonschema/gen.sh
```diff
@@ -0,0 +1,99 @@
...
+echo "- Downloading protobuf tools"
+PROTOBUF_VERSION=3.12.3
+curl -sSL \
+  https://github.com/protocolbuffers/protobuf/releases/download/v$PROTOBUF_VERSION/protoc-$PROTOBUF_VERSION-$PROTOBUF_PLATFORM.zip \
+  --output protobuf.zip
+unzip -q protobuf.zip
+cd ..
+
+echo "- Setting up protoc plugin (chrusty/protoc-gen-jsonschema)"
+go get -v github.com/chrusty/protoc-gen-jsonschema/cmd/protoc-gen-jsonschema
```
grayside, line 44: If you wanted to pin a specific version of the protoc plugin, it would be on this line.

tools/proto2jsonschema/gen.sh
```diff
@@ -0,0 +1,99 @@
...
+for proto in $DATA_PROTOS; do
+  ## For each data.proto file, generate the jsonschema
+  # proto e.g. proto/google/events/cloud/pubsub/v1/data.proto
+
+  # Cut proto/ prefix
+  PATH_PREFIX="proto/"
+  PATH_PREFIX_LEN=$((1 + $(echo ${#PATH_PREFIX})))
+
+  # e.g. google/events/cloud/pubsub/v1/data.proto
+  CUT_PROTO=$(echo "$proto" | cut -c $PATH_PREFIX_LEN-)
```
grayside, line 64: Not even a nit, as this approach is fine. However, if you want a more concise implementation you can use a bash substitution:

```suggestion
  # Remove proto/ prefix
  PATH_PREFIX="proto/"
  CUT_PROTO=${proto#"$PATH_PREFIX"}
```

You can get it down to one line if you inline the PATH_PREFIX variable.

tools/proto2jsonschema/gen.sh
```diff
@@ -0,0 +1,99 @@
...
+
+  # e.g. google/events/cloud/pubsub/v1/
+  PATH_POSTFIX="/data.proto"
+  PATH_POSTFIX_LEN=$((1 + $(echo ${#PATH_POSTFIX})))
+  CUT_PROTO_DIR=$(echo "$CUT_PROTO" | rev | cut -c $PATH_POSTFIX_LEN- | rev)
```
grayside, line 69: Another bash substitution option:

```suggestion
  PATH_POSTFIX="/data.proto"
  CUT_PROTO_DIR=${proto%"$PATH_POSTFIX"}
```

tools/proto2jsonschema/postgen.js
```diff
@@ -0,0 +1,46 @@
...
+const ROOT = path.resolve(`${__dirname}/../../jsonschema`);
+console.log(`Fixing paths in dir: ${ROOT}`);
+(async () => {
+  const filePaths = await recursive(ROOT);
+  // For every file
+  filePaths.map(filePath => {
+    const dataName = path.basename(filePath,  path.extname(filePath)); // i.e. LogEntryData
+    
+    // Create the modified JSON schema output
+    const json = JSON.parse(fs.readFileSync(filePath).toString());
```
grayside, line 22: Might be worth a try/catch on this to point out the protoc step had a problem.

tools/proto2jsonschema/postgen.js
```diff
@@ -0,0 +1,46 @@
...
+    const resultJSON = {
+      // Add the name and $id first
+      name: dataName,
+      $id: getId(filePath),
+      // Add all other fields. Convert keys to camelCase (i.e. remove snake_case keys)
+      ...camelcaseKeys(json, {deep: true})
+    };
+
+    // Write back file with formatting
+    fs.writeFileSync(filePath, JSON.stringify(resultJSON, null, 2));
```
grayside, line 32: Nice touch for human readability.

tools/proto2jsonschema/postgen.js
```diff
@@ -0,0 +1,46 @@
...
+    const dataName = path.basename(filePath,  path.extname(filePath)); // i.e. LogEntryData
+    
+    // Create the modified JSON schema output
+    const json = JSON.parse(fs.readFileSync(filePath).toString());
+    const resultJSON = {
+      // Add the name and $id first
+      name: dataName,
+      $id: getId(filePath),
+      // Add all other fields. Convert keys to camelCase (i.e. remove snake_case keys)
+      ...camelcaseKeys(json, {deep: true})
```
grayside, line 28: nit: Slight performance optimization inserting the name and $id into the output of camelcaseKeys instead of copying the existing schema into this new object.

tools/proto2jsonschema/postgen.js
```diff
@@ -0,0 +1,46 @@
...
+const fs = require('fs');
+const recursive = require("recursive-readdir");
+const camelcaseKeys = require('camelcase-keys');
+
+/**
+ * This tool polishes the JSON schemas with a few modifications:
+ * - Adds "$id" – "." delimited ID – e.g. "google.events.cloud.audit.v1.LogEntryData"
+ * - Adds "name" – The name for the JSON schema – e.g. "LogEntryData"
+ * - Removes snake_case fields that are duplicative of camelCase fields that are seen in JSON payloads
+ *   - This is a feature request for the proto2jsonschema generator.
```
grayside, line 11: nit: Add a TODO(feature request link)

tools/proto2jsonschema/postgen.js
```diff
@@ -0,0 +1,46 @@
+const path = require('path');
+const fs = require('fs');
+const recursive = require("recursive-readdir");
+const camelcaseKeys = require('camelcase-keys');
+
+/**
+ * This tool polishes the JSON schemas with a few modifications:
+ * - Adds "$id" – "." delimited ID – e.g. "google.events.cloud.audit.v1.LogEntryData"
```
grayside, line 8: Minor (nit+1):
Top-level $id is meant to resolve to an absolute URI.
https://json-schema.org/understanding-json-schema/structuring.html#the-id-property
https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.8.2.2

This is technically valid, but it does lack the self-reference to canonical hosting URL which is often used and as of json-schema Draft 6 used as a base url for $ref.

grant:
Will definitely make sure this change is 1 commit that doesn't include any history of the changes/submodule in this PR.

grayside (approved):
Reviewed and have some feedback but nothing I regard as a blocker.

tools/proto2jsonschema/postgen.js
grant: Changed to absolute URI ref per Chat discussion.

tools/proto2jsonschema/postgen.js
grant: ack

tools/proto2jsonschema/postgen.js
grant: There isn't much performance optimization for not copying the JSON file, especially at the sacrifice of readability.

I'm seeing this time for the whole postgen script:

```
real	0m0.107s
user	0m0.108s
sys	0m0.025s
```

I think optimizing should come at >10 seconds, and these <100 fs reads are probably not the first place to optimize.

tools/proto2jsonschema/postgen.js
grant: Good comment, however I think JSON syntax validation after the protoc step should be separate.

If there are FS errors, I think reading those FS errors rather than any caught errors would be easier to debug.

grant:
FYI, I've verified there was no submodule committed with this PR squash when merged into master.

jskeet:
> FYI, I've verified there was no submodule committed with this PR squash when merged into master.

Right - but if you squash it locally into a single commit and force-push, it means we can see the results more easily *before* you merge.

jskeet:
(If you want to just use squash and merge, that's okay - I just generally prefer to have the PR show what commits will end up in the main branch before merging. Just a personal preference.)

PR #105, opened 2020-11-04
docs: add cloudevent product name column

### LINKED ISSUE #67: Add "Name" column to README table
## Expected Behavior

Add a `Source Name` or similar column to the README tab.

When focused on a use-case oriented onboarding to using CloudEvents without a sample, I'm looking for something which says "Audit Logs" or "Cloud Storage", not the existing identifiers.

## Actual Behavior

The machine readable names are useful for developers already oriented on the event types. There is no plain human-readable label.

grant:
Fixes #67

- Adds Product name column
  - These human-readable product name are not structurally defined in protos
  - Exact spelling pulled from product pages
- README table generator throws exception when trying to gen with an unknown package
- Updates comments around column numbers. The ordering of columns is not imperative and can be changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,18 +22,18 @@ may change. Even the core `cloud_event_type` annotation may change.
 
 This repository contains definitions for the following CloudEvents:
 
-|Package|Event types|Data messages|
-|-|-|-|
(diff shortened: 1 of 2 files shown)


jskeet (approved):
I'll probably send another PR to fetch the product name from a dictionary, which will simplify the code a tiny bit, but it's fine to merge this now.

PR #107, opened 2020-11-12
fix: update java headers and ruby

grant:
Fixes Java headers and Ruby file names.

See https://github.com/googleapis/google-cloudevents-java/issues/38


tools/quicktype-wrapper/src/index.ts
```diff
@@ -67,7 +67,7 @@ async function getJSONSchemasPaths(directory: string) {
  * @returns {string} The filename, like document_event_data
  */
 function getFilename(typeName: string, lang: string) {
-  if (lang === 'GOLANG') {
+  if (lang === 'GOLANG' || lang === 'RUBY') {
```
grayside, line 70: A comment along the lines of "Golang and Ruby prefer filenames to use lower snakecase instead of CamelCase" would be helpful...

grayside (approved):

tools/quicktype-wrapper/src/index.ts
grant: Added.

PR #108, opened 2020-11-17
DO NOT MERGE: Sample of multiple events per product

jskeet:
This does not represent a real pair of new events - it's just a
sample of what the protos might look like if Storage were to publish
events about buckets as well as about objects, so we can see what
the effect on the JSON schema (etc) might be.


jskeet:
Created as a regular PR in order to send it out for review, but definitely not intended to be merged!

grant:
Useful, thanks

grant:
Closing the PR as I can still see and use the code. Thanks again.

PR #110, opened 2020-11-17
fix: return of genJSONSchemasAndGenFiles()

michaelawyu:
Fix the return of the `getJSONSchemasAndGenFiles` method.


grant:
Are you sure that we want to get the first key of the genfiles rather than the file contents?

Can you give instructions on what change this makes with some language?

michaelawyu:
@grant Actually it is getting the file content; before this patch the method will throw an error as the `genFile` output is no longer an array after a previous qt tool update. This method is only used in Go and Python library generation.

grant (approved):

PR #112, opened 2020-11-18
feat: enhances jsonschema

grant:
- Enhances CloudEvent JSON schemas with `package`, `datatype`, and `types` fields. Data is from protos as source of truth still.
  - Implements part 1 of 2 for design doc "Proposal: Augment JSON Schemas".
- Updates README.

---

Part 2 will be to create the JSON schema registry.

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -4,7 +4,7 @@ The tools in this directory are not required in order to *use* the
 rest of the content of this repository. They are present to help
 create or maintain the content.
 
-## [readme-list](./readme-list/readme.md)
+## [registry](./registry/readme.md)
 
 This tool edits the top-level [README](../README.md) event list.
 
@@ -15,7 +15,3 @@ This tool wraps the `quicktype` tool to generate any language using JSON schemas
 ## [proto2jsonschema](./proto2jsonschema/README.md)
 
 This tool converts protobufs to JSON schemas.
-
-## [jsonschema-registry](./jsonschema-registry/README.md)
-
-This tool generates a registry for JSON schemas.
--- a/tools/README.md
+++ b/tools/README.md
@@ -1,24 +1,21 @@
-# Tools
+# Google CloudEvent Repo Tools
 
 The tools in this directory are not required in order to *use* the
 rest of the content of this repository. They are present to help
 create or maintain the content.
 
-While they are *currently* written in C#, requiring .NET Core 3.1 to
-build and run, this is purely incidental. Pull requests to rewrite
-the tools in languages more commonly used at Google (e.g. Java, Go
-or Python) are welcome.
+## [readme-list](./readme-list/readme.md)
 
-## registry
+This tool edits the top-level [README](../README.md) event list.
 
-This tool edits the top-level [README](../README.md) to maintain the
-event registry. It requires that the protobuf descriptor set (a
-protobuf representation of all the .proto files) is already
-compiled via protoc. (The [root compile.sh script](../compile.sh)
-does this.)
+## [quicktype-wrapper](./quicktype-wrapper/README.md)
 
-## quicktype-wrapper
+This tool wraps the `quicktype` tool to generate any language using JSON schemas.
 
-This tool wraps the quicktype tool to generate code using JSON Schema.
+## [proto2jsonschema](./proto2jsonschema/README.md)
 
-See [`quicktype-wrapper/`](./quicktype-wrapper/README.md).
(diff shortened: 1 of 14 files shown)


tools/proto2jsonschema/postgen.js
```diff
@@ -120,17 +126,55 @@ function getId(filepath) {
...
+  const protoPath = packageName.split('.').join('/');
+  const eventPath = path.resolve(`${__dirname}/../../proto/${protoPath}/events.proto`);
+  const proto = protobufjs.loadSync(eventPath);
+  const protoAsJSON = proto.toJSON();
+
+  // Flatten the protobuf representation and look directly for these keys:
+  const CLOUD_EVENT_TYPE = '(google.events.cloud_event_type)';
+  const CLOUD_EVENT_PRODUCT = '(google.events.cloud_event_product)';
+  const flattenedProtoValueMap = flatten(protoAsJSON);
+  const flattenedProtoValueMapEntries = Object.entries(flattenedProtoValueMap);
```
jskeet, line 158: Can we not predict the structure of this JSON? It feels a bit odd to just look through every entry in the JSON representation, rather than looking for specific paths (product as a top-level file option, event type as a message option).

I'm okay with it for now, but it doesn't sit quite right.

tools/README.md
```diff
@@ -1,24 +1,21 @@
...
 create or maintain the content.
 
-While they are *currently* written in C#, requiring .NET Core 3.1 to
-build and run, this is purely incidental. Pull requests to rewrite
-the tools in languages more commonly used at Google (e.g. Java, Go
-or Python) are welcome.
+## [readme-list](./readme-list/readme.md)
 
-## registry
+This tool edits the top-level [README](../README.md) event list.
```
jskeet, line 9: Will you send a later PR that removes the C# code then? (To confirm: I don't see any reason to keep it.)

tools/proto2jsonschema/postgen.js
```diff
@@ -120,17 +126,55 @@ function getId(filepath) {
...
+      types.push(v);
+    }
+    if (k.endsWith(CLOUD_EVENT_PRODUCT)) {
+      product = v;
+    }
+  }
+
+  // Return types and product.
+  return {
+    types,
```
jskeet, line 174: Is "types" a clear enough name here? I'm just imagining someone loading the schema file without much context, and expecting "types" to mean data types rather than CloudEvent types. If we were to say "cloudEventTypes" it would potentially be clearer - but clunkier, certainly.

I'm on the fence (and this definitely doesn't block approval) - just something to think about before anything starts depending on the name, making it harder to change.

tools/README.md
jskeet: Ah - I see this change was reverted in the second commit. I don't mind whether the C# code stays (if it's simpler than rewriting it) or goes (if the README editing drops out naturally).

jskeet (approved):

tools/README.md
grant: Yeah, I started reworking the registry based off of our JSON schemas, but then stopped as I didn't want to implement too much in one PR.

Will send a follow-up PR for the registry part.

tools/proto2jsonschema/postgen.js
grant: I think I'll change to `cloudeventTypes` in the JSON schema to make it more clear.

tools/proto2jsonschema/postgen.js
grant: Yeah, I'm not really proud of this solution to extract the cloudevent fields.

With the solution, I wanted to parse the protos (rather than string manipulation).

This is the JSON representation of the proto we get for our events.protos:

```json
{"nested":{"google":{"nested":{"events":{"nested":{"cloud":{"nested":{"audit":{"nested":{"v1":{"options":{"csharp_namespace":"Google.Events.Protobuf.Cloud.Audit.V1","(google.events.cloud_event_product)":"Cloud Audit Logs"},"nested":{"AuditLogWrittenEvent":{"options":{"(google.events.cloud_event_type)":"google.cloud.audit.log.v1.written"},"fields":{"data":{"type":"LogEntryData","id":1}}}}}}}}}}}}}}}
```

Another solution would be to recursively traverse the object and find the key. Since the key is unique, just flattening the map and getting the key sounded fine.

Maybe I should add a comment with an example result from protobufjs's toJSON.

grayside (approved):

Issue #116, opened 2020-11-23
Add CI setup for ensuring JSON schemas, catalog, and readme are updated


grant:
## Expected Behavior

This repo has a few tools that generate files from our proto source of truth. In particular:
- JSON schemas
- JSON schema catalog
- README catalog

In order to ensure that these generated files are current, we should:

- Make it easy to run all generators / tools at once
- Ensure the git diff of a PR is 0 when re-running the generators

GitHub CI should check this on every PR. The CI result should state any errors with exactly which of the three tools did not produce incorrect results.

## Actual Behavior

We don't ensure the generators are all in-sync.

We have a `build.yaml` workflow, but it's a bit out-of date and doesn't state exactly which

PR #115, opened 2020-11-22
README Registry Generator

grant:
Enhances README with generated list of CloudEvents including links to proto/json schemas.

## Preview (a bit outdated)

|Product|Schemas|Types|
|-|-|-|
|Cloud Audit Logs|[Proto](proto/google/events/cloud/audit/v1/data.proto) / [JSON](https://googleapis.github.io/google-cloudevents/jsonschema/google/events/cloud/audit/v1/LogEntryData.json)|<br>Data Type:<br>`google.cloud.audit.v1.LogEntryData`<br>CloudEvent Type(s):<br>`google.cloud.audit.log.v1.written`|
|Cloud Build|[Proto](proto/google/events/cloud/cloudbuild/v1/data.proto) / [JSON](https://googleapis.github.io/google-cloudevents/jsonschema/google/events/cloud/cloudbuild/v1/BuildEventData.json)|<br>Data Type:<br>`google.cloud.cloudbuild.v1.BuildEventData`<br>CloudEvent Type(s):<br>`google.cloud.cloudbuild.build.v1.statusChanged`|
|Cloud Firestore|[Proto](proto/google/events/cloud/firestore/v1/data.proto) / [JSON](https://googleapis.github.io/google-cloudevents/jsonschema/google/events/cloud/firestore/v1/DocumentEventData.json)|<br>Data Type:<br>`google.cloud.firestore.v1.DocumentEventData`<br>CloudEvent Type(s):<br>`google.cloud.firestore.document.v1.created`<br/>`google.cloud.firestore.document.v1.updated`<br/>`google.cloud.firestore.document.v1.deleted`<br/>`google.cloud.firestore.document.v1.written`|
|Cloud Pub/Sub|[Proto](proto/google/events/cloud/pubsub/v1/data.proto) / [JSON](https://googleapis.github.io/google-cloudevents/jsonschema/google/events/cloud/pubsub/v1/MessagePublishedData.json)|<br>Data Type:<br>`google.cloud.pubsub.v1.MessagePublishedData`<br>CloudEvent Type(s):<br>`google.cloud.pubsub.topic.v1.messagePublished`|
[...]

PR #117, opened 2020-11-23
ci: add CI to run through all build tools

grant:
## ci: add CI to run through all build tools

- Runs jsonschema/registry tools on CI. Ensures running the tools produce a 0 diff.
- Splits schema validator workflow with build workflow.

---

- Requires: https://github.com/googleapis/google-cloudevents/pull/115 (Note: diff will be a bit smaller once that PR is merged.)
- Fixes: https://github.com/googleapis/google-cloudevents/issues/116


grant:
Hey @jskeet, @grayside, this PR should be ready.

It will ensure that the tools are run whenever we make a change to a proto or tool itself. Also makes the order of running these tools clear.

jsonschema/catalog.json
```diff
@@ -21,12 +12,15 @@
       ]
     },
     {
-      "url": "https://googleapis.github.io/google-cloudevents/jsonschema/google/events/cloud/cloudbuild/v1/BuildEventData.json",
-      "name": "BuildEventData",
-      "description": "Build event data\n Common build format for Google Cloud Platform API operations.\n Copied from\n https://github.com/googleapis/googleapis/blob/master/google/devtools/cloudbuild/v1/cloudbuild.proto.",
-      "datatype": "google.events.cloud.cloudbuild.v1.BuildEventData",
+      "url": "https://googleapis.github.io/google-cloudevents/jsonschema/google/events/cloud/firestore/v1/DocumentEventData.json",
```
jskeet, line 15: Why are all these changes here? Is the ordering non-deterministic at the moment, or was it non-deterministic before and this just corrects it?

.github/workflows/generate.yaml
```diff
@@ -0,0 +1,25 @@
+name: Generate
```
jskeet, line 1: Will this (and Validate) be run even if Compile fails? Can we make it wait until Compile succeeds before this is run?

(This is why I prefer them being separate steps in one action - while the PR description says it "Also makes the order of running these tools clear" it feels to me like it's doing the opposite.)

jsonschema/catalog.json
grant: It's currently non-deterministic. I state that the PR will be smaller due to this issue in the PR description.

I need to merge https://github.com/googleapis/google-cloudevents/pull/115 before this PR to not see this artifact.

.github/workflows/generate.yaml
grant: I can merge them in one large action, although I would rather see 1/2 checks pass or 2/2 checks pass than 0/1 checks pass in the UI. Yes, the 2nd workflow (generate) cannot pass if the 1st workflow fails (compile).

The order is clearer as we have 4 tools in https://github.com/googleapis/google-cloudevents/tree/master/tools and we do not describe the specific order required to run them to correctly generate results. This CI does that, as you can just look at the CI and run the commands rather than looking at each tool.

jsonschema/catalog.json
jskeet: Ah, right. I'd missed that.

.github/workflows/generate.yaml
jskeet: I think it's subjective, and I guess you'll be working in this repo more than I will be. I think the order is clearest if there's just *one* action that takes all the steps, but it's your call.

jskeet (approved):
Approved contingent on #115 being merged, etc.

Issue #118, opened 2020-11-25
Data Proto Descriptions Are Inconsistent – CAL, Cloud Build, Firebase Auth


grant:
## Expected Behavior

Data proto messages have similar formats and simple descriptions.

Example (good): 

https://github.com/googleapis/google-cloudevents/blob/master/proto/google/events/cloud/pubsub/v1/data.proto#L21-L22

```proto
// The data received in an event when a message is published to a topic.
message MessagePublishedData {
```

## Actual Behavior

Different have different description formats.

- Specifically, **CAL** and **Cloud Build** protos have different descriptions than the others.
- nit: **Firebase Auth** description does not have a period unlike all other protos.

## Example – Cloud Build

https://github.com/googleapis/google-cloudevents/blob/master/proto/google/events/cloud/cloudbuild/v1/data.proto#L24-L28

### Expected

```proto
// Build event data.
message BuildEventData {
```

### Actual

```proto
// Build event data
// Common build format for Google Cloud Platform API operations.
// Copied from
// https://github.com/googleapis/googleapis/blob/master/google/devtools/cloudbuild/v1/cloudbuild.proto.
message BuildEventData {
```

This is causing the schemas/doc generator to produce not ideal results:

# Example Doc Gen Results

### Cloud Audit Logs

```js
const {toLogEntryData} = require('@google/events/cloud/audit/v1/LogEntryData');

/**
 * Generic log entry, used as a wrapper for Cloud Audit Logs in events.
 This is copied from
 https://github.com/googleapis/googleapis/blob/master/google/logging/v2/log_entry.proto
 and adapted appropriately.
 */
const data = {
  // ...
};

const jsExample = toLogEntryData(data);
console.log(jsExample);
```

### Cloud Build

```js
const {toBuildEventData} = require('@google/events/cloud/cloudbuild/v1/BuildEventData');

/**
 * Build event data
 Common build format for Google Cloud Platform API operations.
 Copied from
 https://github.com/googleapis/googleapis/blob/master/google/devtools/cloudbuild/v1/cloudbuild.proto.
 */
const data = {
  // ...
};

const jsExample = toBuildEventData(data);
console.log(jsExample);
```

### Cloud Firestore

```js
const {toDocumentEventData} = require('@google/events/cloud/firestore/v1/DocumentEventData');

/**
 * The data within all Firestore document events.
 */
const data = {
  // ...
};

const jsExample = toDocumentEventData(data);
console.log(jsExample);
```

[...]

PR #120, opened 2020-11-30
fix: standardize data proto descriptions

grant:
Reviews all data.proto descriptions. These descriptions are used in the JSON schemas and client libraries.

Fixes https://github.com/googleapis/google-cloudevents/issues/118


jskeet (approved):
Yup, I'm fine with all of these. As well as republishing "downstream" to JSON schema etc, we'll need to republish "upstream" to internal source control as well. (I would suggest making the change there and letting it propagate, but we haven't got that configured yet as far as I know.)

PR #123, opened 2020-12-02
fix: remove oneof type for 64 bit numbers

grant:
Uses `type: string` for proto `int64` (and related (`*64`) types per [proto JSON representation spec](https://developers.google.com/protocol-buffers/docs/proto3#json).

This is causing an issue in the golang from jsonschema generator: https://github.com/googleapis/google-cloudevents-go/issues/22

---

I used the master branch of the `protoc-gen-jsonschema` tool to get this fix.

See issue: https://github.com/chrusty/protoc-gen-jsonschema/issues/47

### DIFF
--- a/jsonschema/google/events/cloud/audit/v1/LogEntryData.json
+++ b/jsonschema/google/events/cloud/audit/v1/LogEntryData.json
@@ -88,14 +88,7 @@
           "description": "The resource's original state before mutation. Present only for\n operations which have successfully modified the targeted resource(s).\n In general, this field should contain all changed fields, except those\n that are already been included in `request`, `response`, `metadata` or\n `service_data` fields.\n When the JSON object represented here has a proto equivalent,\n the proto name will be indicated in the `@type` property."
         },
         "numResponseItems": {
-          "oneOf": [
-            {
-              "type": "integer"
-            },
-            {
-              "type": "string"
-            }
-          ],
+          "type": "string",
           "description": "The number of items returned from a List or Query API method,\n if applicable."
         },
         "status": {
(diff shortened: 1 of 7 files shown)


jskeet (commented):
Is this a manual change to the JSON schema, or have you changed the tooling and this is the result of that change? I think we should keep the schema to just the results of generation.

(As a guide to this sort of change, the PR format I find to be useful is a first commit that changes the generator, then a second commit that runs the generator. YMMV, but that's worked well for us in .NET client libraries land.)

The change itself looks fine btw :)

grant:
This is re-running the generator with the latest protoc plugin, fix for https://github.com/chrusty/protoc-gen-jsonschema/issues/47#issuecomment-736741974.

The fix is on `master` but Crusty still needs to cut a release tag.

> Is this a manual change to the JSON schema, or have you changed the tooling and this is the result of that change? I think we should keep the schema to just the results of generation.
> 
> (As a guide to this sort of change, the PR format I find to be useful is a first commit that changes the generator, then a second commit that runs the generator. YMMV, but that's worked well for us in .NET client libraries land.)

This is re-running the generator with the latest protoc plugin, fix for https://github.com/chrusty/protoc-gen-jsonschema/issues/47#issuecomment-736741974.

The fix is on `master` but Crusty still needs to cut a release tag for 0.9.7 or 0.10.0. https://github.com/chrusty/protoc-gen-jsonschema/tags

jskeet:
I see. If this is blocking anything (which I can imagine it might be) then we could potentially merge it - ideally with an issue to keep us honest, as it were. Otherwise, we could keep this PR open until there's a tag. Let me know what your thinking is.

grant:
I mean, nobody is cutting a release for our protobufs. But I can wait a bit for that repo to cut a release.

grant:
Waiting sounds good.

grant:
@jskeet @grayside The proto plug-in version has been updated, so this PR should be ready to merge. PTAL.

jskeet (approved):

PR #124, opened 2020-12-07
Modifies the CloudEvent JSON schema to Have Accurate Field Names

grant:
Updates the CloudEvent JSON schema to have accurate field names.

- The field `cloudevent` isn't accurate here as this is not the ID for the CloudEvent (it doesn't include the name).
- The field `package` is the package in which this event lives, excluding the `name`.
- The field `datatype` is the full type of the `data` that this schema represents. = `package` + `name`.


grant:
Oops, this is an old branch. Closing.

Issue #125, opened 2020-12-09
JSON Schema: Use Type Integer Rather Than String


grant:
## Expected Behavior

The JSON schemas should use type integer for `int64` types, like with `port` in CAL: https://github.com/googleapis/google-cloudevents/blob/master/jsonschema/google/events/cloud/audit/v1/LogEntryData.json#L420

This was mentioned by @jskeet here: https://github.com/googleapis/google-cloudevents-java/pull/54#discussion_r538083491

## Actual Behavior

We use type `string`, which causes end libraries to not have great results.

Issue #126, opened 2020-12-11
JSON Schema Gen'd Examples are Empty


grant:
## Expected Behavior

When running the json schema generator from master, there should be no changes in the generated files:

```sh
./scripts/gen.sh
```

## Actual Behavior

Running the generator creates `examples` fields that are empty (`[]`).

```
+  "examples": [],
```

This is introduced here: https://github.com/googleapis/google-cloudevents/pull/111

michaelawyu:
A PR (#130) has been sent.

grant:
> A PR (#130) has been sent.

Thanks. I added a comment, and we should be able to merge after it's resolved.

grant:
This was fixed in #130.

PR #127, opened 2020-12-11
JSON schema no bigint strings

grant:
Fixes https://github.com/googleapis/google-cloudevents/issues/125

- For JSON schemas, uses `integer` instead of `string` for integer values.
  - For example CAL `port` field.
- Change in `examples` is due to a bug recorded here: https://github.com/googleapis/google-cloudevents/issues/126
  - Hopefully we can fix that before merging this PR.


jskeet (commented):
This appears to capture the fact that it's an integer (good) but not the size... I can't see how the Java generator would generate int/Integer or long/Long properties based on this.

grant:
To see how this change causes the Java generator to use a `Long` instead of a `string`, copy the updated JSON schema:

https://raw.githubusercontent.com/googleapis/google-cloudevents/grant_jsonschema_no_bigint_strings/jsonschema/google/events/cloud/audit/v1/LogEntryData.json

Paste it in `https://app.quicktype.io/`, ensuring the source type is JSON schema and language is Java (or Go).

The result is that this field is a `Long`/`int64`.

## Example

```java
/**
 * The log entry payload, which is always an AuditLog for Cloud Audit Log events.
 */
public class ProtoPayload {
    private AuthenticationInfo authenticationInfo;
    private AuthorizationInfo[] authorizationInfo;
    private Metadata metadata;
    private String methodName;
    private Long numResponseItems;  <---  Long!
    private Request request;
    private RequestMetadata requestMetadata;
    private ResourceLocation resourceLocation;
    private String resourceName;
    private ResourceOriginalState resourceOriginalState;
    private Response response;
    private ServiceData serviceData;
    private String serviceName;
    private Status status;
```

Ready for review.

jskeet (approved):
Okay, so this change *only* affects int64 fields... but it looks like we're using "integer" for int32 fields as well, e.g. for `StorageObjectData.component_count`. So that's still a problem IMO - just a different one to the one that's fixed here.

(In general, my main concern over the JSON schema is that it's losing a significant amount of type information.)

grant:
Yes, I believe this only affects *64 proto fields.

Proto vs JSON schema based libraries is a language-per-language decision. If there are known caveats using the JSON schema description, such as int32 described above, that affects clients significantly, we should evaluate that per language.

grant:
Merging this string int fix such that we can get on to other PRs.

PR #130, opened 2020-12-15
fix: patch postgen.js to add examples field back

michaelawyu:
Note: `postgen.js` will fail if `catalog.json` is present in the `jsonschema` directory.

### DIFF
--- a/tools/proto2jsonschema/postgen.js
+++ b/tools/proto2jsonschema/postgen.js
@@ -165,7 +165,7 @@ function getExamples(filepath) {
       const isJSONFile = value.isFile() && value.name.endsWith('.json');
       return isJSONFile ? `https://googleapis.github.io/google-cloudevents/testdata/${removeSuffix}/${value.name}` : null;
     })
-    .filter((value) => !value);
+    .filter((value) => value != null);
 }
 
 /**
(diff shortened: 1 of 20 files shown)


grant:
Can we first make the CI pass (maybe a different PR), then submit this PR?

tools/proto2jsonschema/postgen.js
```diff
@@ -165,7 +165,7 @@ function getExamples(filepath) {
       const isJSONFile = value.isFile() && value.name.endsWith('.json');
       return isJSONFile ? `https://googleapis.github.io/google-cloudevents/testdata/${removeSuffix}/${value.name}` : null;
     })
-    .filter((value) => !value);
+    .filter((value) => value != null);
```
grant, line 168:
```ts
 .filter((value) => value !== null);
```

grant (approved):

grant:
I suggest we fix the CI first, get that passing, so we can ensure this PR fixes the error.

https://github.com/googleapis/google-cloudevents/pull/132

WDYT?

grant:
Hi @michaelawyu, is this PR still being worked on?

grant:
CI passed, and PR is approved, so I merged.

PR #132, opened 2020-12-17
fix: Fix CI. Regen with latest Chrusty plugin.

grant:
Now that we have CI checking that the full generation steps is working, this PR makes the CI pass (through a few adjustments).

- Ensures the chrusty protoc plugin that is specified is fixed
  - Ensures anyone running this locally doesn't use a cached version by deleting the cache beforehand.
- Latest crusty plugin has some updates:
  - Fixes bad `oneOf` types allowing nulls. https://github.com/chrusty/protoc-gen-jsonschema/commit/7726bafee96859cdee2947f818544b2af809be75
    - This is a good thing as having it causes complicated behavior in language clients.
  - Fixes enum values per: https://github.com/chrusty/protoc-gen-jsonschema/pull/44
- Slightly adjust output of generator scripts to be more readable and debuggable.
- Adds cleaning script for helping the system packages are correct.

You'll notice CI passes now. Fixes https://github.com/googleapis/google-cloudevents/issues/140.

### DIFF
--- a/jsonschema/google/events/cloud/audit/v1/LogEntryData.json
+++ b/jsonschema/google/events/cloud/audit/v1/LogEntryData.json
@@ -71,14 +71,8 @@
           "properties": {
             "fields": {
               "additionalProperties": {
-                "oneOf": [
-                  {
-                    "type": "null"
-                  },
-                  {
-                    "type": "object"
-                  }
-                ]
+                "additionalProperties": true,
+                "type": "object"
               },
               "type": "object",
               "description": "Unordered map of dynamically typed values."
@@ -140,14 +134,8 @@
               "properties": {
                 "fields": {
                   "additionalProperties": {
-                    "oneOf": [
-                      {
-                        "type": "null"
-                      },
-                      {
-                        "type": "object"
-                      }
-                    ]
+                    "additionalProperties": true,
+                    "type": "object"
                   },
                   "type": "object",
                   "description": "Unordered map of dynamically typed values."
@@ -174,14 +162,8 @@
                         "properties": {
                           "fields": {
                             "additionalProperties": {
-                              "oneOf": [
-                                {
-                                  "type": "null"
-                                },
-                                {
-                                  "type": "object"
-                                }
-                              ]
+                              "additionalProperties": true,
+                              "type": "object"
                             },
                             "type": "object",
                             "description": "Unordered map of dynamically typed values."
@@ -202,14 +184,8 @@
                         "properties": {
                           "fields": {
                             "additionalProperties": {
-                              "oneOf": [
-                                {
-                                  "type": "null"
-                                },
-                                {
-                                  "type": "object"
-                                }
-                              ]
+                              "additionalProperties": true,
+                              "type": "object"
                             },
                             "type": "object",
                             "description": "Unordered map of dynamically typed values."
@@ -375,14 +351,8 @@
                       "properties": {
                         "fields": {
                           "additionalProperties": {
-                            "oneOf": [
-                              {
-                                "type": "null"
-                              },
-                              {
-                                "type": "object"
-                              }
-                            ]
+                            "additionalProperties": true,
+                            "type": "object"
                           },
                           "type": "object",
                           "description": "Unordered map of dynamically typed values."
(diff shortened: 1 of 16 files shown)


jskeet:
I'm now on vacation until Jan 5th - hopefully one of the other reviewers is still working :)

grant:
> I'm now on vacation until Jan 5th - hopefully one of the other reviewers is still working :)

No worries. I'll be on vacation shortly too. I might stack up some PRs that don't have urgency.

jsonschema/google/events/cloud/audit/v1/LogEntryData.json
```diff
@@ -1,6 +1,7 @@
 {
```
grayside, line 1: I'm not sure if there's a better way, but when reviewing process/automation changes, these generated output changes tend to become easily ignored noise.

Is it intended that these diffs be reviewed as well? If so it might be useful to deep link or embed some signature diff sections in the PR description.

scripts/gen.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -e
 
+# Clean directory and cache
+go clean -cache
+rm -rf $GOPATH/src/github.com/
```
grayside, line 6: I appreciate this work to create a deterministic build environment, however this seems pretty disruptive if run locally. Is this intended for CI? Maybe a cleaner approach would be to include a go.mod file?

tools/proto2jsonschema/gen.sh
```diff
@@ -29,21 +29,22 @@ case "$OSTYPE" in
...
-PROTOBUF_VERSION=3.12.3
+echo "- Downloading protobuf@$PROTOBUF_VERSION"
 curl -sSL \
   https://github.com/protocolbuffers/protobuf/releases/download/v$PROTOBUF_VERSION/protoc-$PROTOBUF_VERSION-$PROTOBUF_PLATFORM.zip \
   --output protobuf.zip
 unzip -q protobuf.zip
 cd ..
 
-echo "- Setting up protoc plugin (chrusty/protoc-gen-jsonschema)"
+PROTOC_PLUGING_VERSION=0.9.7
```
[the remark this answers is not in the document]
grant: Fixed.

jsonschema/google/events/cloud/audit/v1/LogEntryData.json
grant: I wrote a detailed description for the PR, covering the `oneOf` and `enum` changes described in this file.

I intend for the review to look at the diff and approve if there are no requested changes. If the PR is not clear, please request for more information in a comment or we can also meet and sync.

jskeet (commented):
I'm basically fine with this but @grayside's comment around the cache cleaning seems reasonable - I don't know enough Go to comment on that.

In terms of making reviews as simple as possible, a technique I've employed in various similar situations is to make sure each commit is either purely manual or purely generated. (So make the script change, commit, regenerate, commit.) I don't think we need to be particularly picky about it - I'm not suggesting splitting up this PR further - but it does make reviews easier.

scripts/gen.sh
grant: I've removed this part of the script from the generator script and added it into a separate script.

I've also filed https://github.com/googleapis/google-cloudevents/issues/141, where we can add a mod file.

Hope this resolves the concern!

grant:
> I'm basically fine with this but @grayside's comment around the cache cleaning seems reasonable - I don't know enough Go to comment on that.
> 
> In terms of making reviews as simple as possible, a technique I've employed in various similar situations is to make sure each commit is either purely manual or purely generated. (So make the script change, commit, regenerate, commit.) I don't think we need to be particularly picky about it - I'm not suggesting splitting up this PR further - but it does make reviews easier.

- I've addressed @grayside's comment around caching.
- In general, yes, that separation of concern is reasonable, although we trade off requiring the reviewer to look at each commit within a PR or losing history if we squash to two commits (manual/gen).

PTAL

jskeet:
I'd still prefer to wait for @grayside to confirm. But yes, I tend to think PRs that deserve multiple commits are best reviewed one commit at a time, and those commits then preserved in the history (not squashed).

grayside (approved):
The follow-up to use go.mod addresses my main concern with this change. I also checked to make sure the clean.sh script isn't directly used by the other scripts.

The removal of "optional null" from a schema is a compatibility break because it means previously valid data structures might now be considered invalid, but as far as I know we don't make any compatibility promises in this repo.

PR #153, opened 2021-02-09
Document unexpected fields, and add PubSub publish_time

jskeet:


proto/README.md
```diff
@@ -118,3 +118,21 @@ code to handle all events for a single resource, without creating
 unnecessary churn when only an unrelated resource type needs
 a new version. (If your code only consumes `bucket.v1` events, the
 introduction of `object.v2` events should not affect you.)
+
+## Unexpected fields
+
+CloudEvent providers may include additional data fields beyond those
+documented here. In particular:
+
+- Some events may include fields with legacy names such as
```
grant, line 127: Define `legacy names`.

I believe you're calling out that the field uses snake_case instead of camelCase for `message_id` here.

proto/README.md
```diff
@@ -118,3 +118,21 @@ code to handle all events for a single resource, without creating
...
+  `message_id` for compatibility purposes.
+- Event providers using proto3 JSON formatting of protobuf `Any`
+  messages may include a field called `@type` as a natural artifact
+  of that formatting process.
+
+Neither of these should concern consumers: it's always possible that
+new fields will have been added to CloudEvents between the time at
+which the code was built and deployed, and the time at which the
+CloudEvent is received. Consumers must therefore handle unexpected
+data fields. This note is primarily to expain some likely causes of
```
grant, line 137: Why "data fields" and not just "fields".

grant (changes requested):
LGTM though there are requested changes/comments.

proto/README.md
jskeet: To make it clear that I'm talking about the data part of the CloudEvent, not context attributes.

proto/README.md
jskeet: I mean names used in the past, that aren't used now.

While the *specific* example we're aware of is using snake_case instead of camelCase, I'm trying to be general. I wouldn't be entirely surprised if we had a similar situation in the future in slightly different ways, and I didn't want to have to rewrite it at that point.

grant (approved):

Issue #176, opened 2021-03-06
proto2jsonschema incorrect field capitalization for fields with numbers


grant:
In the proto2jsonschema step, it looks like some fields are incorrectly capitalized.

## Expected Behavior

`string crc32c = 11;`
->
`"crc32c": { "type": "string" },`

## Actual Behavior

`"crc32C": { "type": "string" },`

With a capital "C".

## Links

Proto:

https://github.com/googleapis/google-cloudevents/blob/master/proto/google/events/cloud/storage/v1/data.proto#L78

JSON Schema:

https://github.com/googleapis/google-cloudevents/blob/master/jsonschema/google/events/cloud/storage/v1/StorageObjectData.json#L56

---

This causes issues with downstream libraries.

Will need to contact chrusty.

grant:
This is due to a bad postgen field capitalization step.

grant:
See above PR for a fix.

PR #178, opened 2021-03-08
JSON Schema postgen native json fieldnames

grant:
Fixes https://github.com/googleapis/google-cloudevents/issues/176

- Uses protoc plugin `json_fieldnames` config, which natively produces JSON fields in the JSON schema rather than a postgen step
  - Primarily notice field `crc32C` is now `crc32c` (which is correct)
- Removes postgen code around changing fields to lowerCamelCase
- `id` changes are unrelated/tangential (we should probably try to remove them in the future)

This will fix https://github.com/googleapis/google-cloudevents-java/issues/58 when the Java library is re-generated.


README.md
```diff
@@ -27,7 +27,7 @@ This repository contains definitions for the following CloudEvents:
...
-|Cloud Firestore|[Proto](proto/google/events/cloud/firestore/v1/data.proto) / [JSON](https://googleapis.github.io/google-cloudevents/jsonschema/google/events/cloud/firestore/v1/DocumentEventData.json)|<br>Data Type:<br>`google.events.cloud.firestore.v1.DocumentEventData`<br>CloudEvent Type(s):<br>`google.cloud.firestore.document.v1.created`<br/>`google.cloud.firestore.document.v1.updated`<br/>`google.cloud.firestore.document.v1.deleted`<br/>`google.cloud.firestore.document.v1.written`|
+|Cloud Firestore|[Proto](proto/google/Docume/data.proto) / [JSON](https://googleapis.github.io/google-cloudevents/jsonschema/google/Docume/DocumentEventData.json)|<br>Data Type:<br>`DocumentEventData`<br>CloudEvent Type(s):<br>`google.cloud.firestore.document.v1.created`<br/>`google.cloud.firestore.document.v1.updated`<br/>`google.cloud.firestore.document.v1.deleted`<br/>`google.cloud.firestore.document.v1.written`|
```
jskeet: This doesn't look right

tools/proto2jsonschema/postgen.js
```diff
@@ -84,46 +79,41 @@ console.log(`Fixing paths in dir: ${ROOT}`);
...
      * This couldn't be done in the previous step, because the $ref was simply wrong in that step.
      * Now the $refs are correct (but long).
      *
      * We need to change the $ref name and definition.
      *
      * We do so by:
-     * - traversing the json, getting all refs
-     * - getting the simplified replacement $ref without the prefix
-     * - replaceAll ref ids (fixes "definition" and corresponding "$ref" tags)
+     * - Looking at the JSON schema "definitions".
```
jskeet, line 90: This seems prone to introducing naming collisions. For example, if there are two references, one to "google.protobuf.Value" and one to "google.events.cloud.firestore.v1.Value", won't those both end up as "Value"?

README.md
grant: This comment is on an outdated diff. Please review the final diff.

tools/proto2jsonschema/postgen.js
grant: I thought about this. TLDR it's literally impossible for a collision unless the proto has a collision `message`.

JSON schema field naming doesn't work like proto naming.

JSON schema has a set of "definitions" and "$ref". All schema definitions are directly in the JSON Schema (at least in our design), no external definitions. The definitions are literally the messages in the proto, so they are "Document" and "Value" in Firestore:

https://github.com/googleapis/google-cloudevents/blob/master/proto/google/events/cloud/firestore/v1/data.proto

`google.protobuf.Value` isn't a JSON Schema definition, and should never be. Only the messages within the data proto can be a JSON Schema definition (if re-used).

---

This is not a new design, it's just implementing the existing design after updating the protoc plugin.

jskeet: So how are references to other messages handled? Do they end up as "unnamed" schema? (I'm guessing that's the case, looking at the Cloud Audit LogEntryData JSON schema.)

(I'm still not entirely sure about this design given that it means two uses of the same "external" proto presumably aren't represented by a common class within the generated library, but that's a separate matter.)

jskeet (approved):
If naming collisions aren't a problem, I'm happy with this. (As noted in comments, I'm still not fully comfortable with the choices that lead to that situation, but that's separate and doesn't need to be addressed in this PR.)

tools/proto2jsonschema/postgen.js
grant: References to other messages are handled via `$ref` fields. I don't know what you're referring to w.r.t. unnamed messages.

I think a new GitHub issue is the best way to address the concern. I'm not sure if you're talking about cross-schema references or some other issue.

jskeet: I've filed https://github.com/googleapis/google-cloudevents/issues/180, describing the concern and how I propose to test it. I'll get to that when I've got schema generation working locally.

Issue #154, opened 2021-02-09
Generation scripts fail on Windows


jskeet:
## Expected Behavior

Running `./scripts/gen.sh` on Windows should regenerate everything appropriately

## Actual Behavior

The schema is regenerated with very different content, and then there's a 

## Steps to Reproduce the Problem

- Start a Windows VM
- Install Go, Node and bash (simplest option for the latter is to use the bash that comes with Git for Windows)
- Clone this repo
- Run `./scripts/gen.sh` from bash and wait for the errors

Error seen:

```text
Error: ENOENT: no such file or directory, open 'C:\users\jonskeet\GitHub\googleapis\google-cloudevents\tools\readme-catalog/../../jsonschema/catalog.json'
    at Object.openSync (fs.js:476:3)
    at Object.readFileSync (fs.js:377:35)
    at Object.<anonymous> (C:\users\jonskeet\GitHub\googleapis\google-cloudevents\tools\readme-catalog\index.js:6:31)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
    at internal/main/run_main_module.js:17:47 {
  errno: -4058,
  syscall: 'open',
  code: 'ENOENT',
  path: 'C:\\users\\jonskeet\\GitHub\\googleapis\\google-cloudevents\\tools\\readme-catalog/../../jsonschema/catalog.json'
}
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @ start: `node .`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the @ start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm WARN Local package.json exists, but node_modules missing, did you mean to install?
```

... but some part of the script had earlier *deleted* `jsonschema/catalog.json`. I don't know why yet - I haven't tried to debug it.

grant:
I need to fix the backslash Windows path separators in this script:

https://github.com/googleapis/google-cloudevents/blob/master/tools/readme-catalog/index.js

Let me know if there's urgency to get to this, or I'll eventually get to it. It shouldn't be blocking core generation.

[...]

PR #179, opened 2021-03-09
ci: run CI on windows

grant:
Runs generate CI on Windows to ensure generation works on that OS too.

See: https://github.com/googleapis/google-cloudevents/issues/154

I can't see a difference on Windows with the generation script, so I'm not sure exactly how to reproduce the issue. Unless, the reported issue is on a older version of Windows or version different than `windows-2019`.

### DIFF
--- a/.github/workflows/generate.yaml
+++ b/.github/workflows/generate.yaml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04, windows-2019]
+        os: [windows-2019, ubuntu-18.04]
     steps:
     - name: Check out repo
       uses: actions/checkout@v2
--- a/.github/workflows/generate.yaml
+++ b/.github/workflows/generate.yaml
@@ -16,7 +16,10 @@ name: Generate
 on: [push, pull_request]
 jobs:
   build:
-    runs-on: [self-hosted, ubuntu-18.04, windows-2019]
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


jskeet (approved):
I'm going to try rerunning the script on my laptop to see whether I'm still seeing an issue - it's possible that it's been silently fixed by another of the changes. It doesn't hurt to keep running it on both platforms though.

jskeet (changes requested):
Having had a look at the output, I'm not convinced this is actually working. 

The output for the generation step on Windows is just:

```text
Run ./scripts/gen.sh
  ./scripts/gen.sh
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
  env:
    GOROOT: C:\hostedtoolcache\windows\go\1.14.15\x64
```

No sign of it actually generating anything - compare that with the Ubuntu output where everything is genuinely running. I suspect the problem is that it's trying to run the script using PowerShell instead of bash.

grant:
> Having had a look at the output, I'm not convinced this is actually working.
> 
> The output for the generation step on Windows is just:
> 
> ```
> Run ./scripts/gen.sh
>   ./scripts/gen.sh
>   shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
>   env:
>     GOROOT: C:\hostedtoolcache\windows\go\1.14.15\x64
> ```
> 
> No sign of it actually generating anything - compare that with the Ubuntu output where everything is genuinely running. I suspect the problem is that it's trying to run the script using PowerShell instead of bash.

Yes. I don't understand Windows well enough or have a Windows machine to debug this.

I'm going to close this PR and keep the issue open.

PR #182, opened 2021-03-11
fix: Fix generation on Windows

### LINKED ISSUE #154: Generation scripts fail on Windows

jskeet:
Fixes #154

### DIFF
--- a/tools/proto2jsonschema/postgen.js
+++ b/tools/proto2jsonschema/postgen.js
@@ -38,6 +38,11 @@ console.log(`Fixing paths in dir: ${ROOT}`);
     
     // Create the modified JSON schema output
     const json = JSON.parse(fs.readFileSync(filePath).toString());
+
+    // Replace backslashes with forward-slashes to allow the script
+    // to work on Windows.
+    filePath = filePath.replace(/\\/g, '/');
+
     const packageName = getCloudEventPackage(filePath);
 
     const resultJSON = {


jskeet:
The other problems I was running into appear to be due to using an old version of the Chrusty plugin. The generation script attempts to specify which version to use, so I'm not quite sure what was going on there, but after removing every trace of it from my Go installation and rerunning, I got to "no changes".

tools/proto2jsonschema/postgen.js
```diff
@@ -38,6 +38,11 @@ console.log(`Fixing paths in dir: ${ROOT}`);
     
     // Create the modified JSON schema output
     const json = JSON.parse(fs.readFileSync(filePath).toString());
+
+    // Replace backslashes with forward-slashes to allow the script
+    // to work on Windows.
+    filePath = filePath.replace(/\\/g, '/');
```
grant, line 44: Great.

I think we want to normalize the file path right when we get it.

Can we do the following:

```js
filePaths.map(filePath => {
  // Normalize path slashes (`/`, `\`) to allow the script to work on Windows.
  filePath = path.normalize(filePath);
```

This replace may work just as well if you prefer it.

https://nodejs.org/api/path.html#path_path_normalize_path

grant (changes requested):
Thanks! 1 comment.

tools/proto2jsonschema/postgen.js
jskeet: I'll give it a try.

jskeet: Nope, that doesn't work - because it still leaves backslashes as backslashes, and later code assumes it's got forward slashes. (Basically it normalizes to "the platform default", rather than normalizing to forward slashes.)

I suspect if we used path operations everywhere instead of using string operations, we'd be okay - but that would be a bigger change.

grant: Alright. I was hoping there's a way to fix the slashes without character replacements, but this seems fine.

grant (approved):

PR #186, opened 2021-03-18
ci: jsonschema validation

grant:
Validates that the JSON schemas are valid schemas according to the JSON schema schema.

I've ran a few CI runs to ensure that this check is correct.

We already have CI that checks for a 0 diff between proto and jsonschema, so naturally that check also breaks when if modifying the json schema manually.

## Example CI runs

Purposeful invalid JSON schema break:

Screen Shot 2021-03-18 at 13 12 25

Purposeful valid JSON schema fix:

Screen Shot 2021-03-18 at 13 12 43

## Notes

Local testing is pretty easy, just install Node and copy the npx script.

### Example CI output

```md
Run JSON_SCHEMAS=$(find ./jsonschema/google/events -name "*.json")
npx: installed 41 in 4.585s
./jsonschema/google/events/cloud/firestore/v1/DocumentEventData.json valid
npx: installed 41 in 2.189s
./jsonschema/google/events/cloud/storage/v1/StorageObjectData.json valid
npx: installed 41 in 2.085s
./jsonschema/google/events/cloud/cloudbuild/v1/BuildEventData.json valid
npx: installed 41 in 2.149s
./jsonschema/google/events/cloud/audit/v1/LogEntryData.json valid
npx: installed 41 in 2.085s
./jsonschema/google/events/cloud/scheduler/v1/SchedulerJobData.json valid
npx: installed 41 in 2.181s
./jsonschema/google/events/cloud/pubsub/v1/MessagePublishedData.json valid
npx: installed 41 in 2.166s
./jsonschema/google/events/firebase/database/v1/ReferenceEventData.json valid
npx: installed 41 in 2.094s
./jsonschema/google/events/firebase/analytics/v1/AnalyticsLogData.json valid
npx: installed 41 in 2.091s
./jsonschema/google/events/firebase/remoteconfig/v1/RemoteConfigEventData.json valid
npx: installed 41 in 2.161s
./jsonschema/google/events/firebase/auth/v1/AuthEventData.json valid
```

### DIFF
--- a/.github/workflows/validate_jsonschemas.yaml
+++ b/.github/workflows/validate_jsonschemas.yaml
@@ -20,7 +20,7 @@ jobs:
     steps:
     - name: Check out repo
       uses: actions/checkout@v2
-    - name: Setup Node for gen/postgen scripts
+    - name: Setup Node
       uses: actions/setup-node@v2
       with:
         node-version: '12'
--- a/.github/workflows/validate_jsonschemas.yaml
+++ b/.github/workflows/validate_jsonschemas.yaml
@@ -26,7 +26,7 @@ jobs:
         node-version: '12'
     - name: Install JSON schema schema
       run: |
-        curl -O http://json-schema.org/draft-04/schema > jsonschema-schema.json
+        curl -o jsonschema-schema.json http://json-schema.org/draft-04/schema
     - name: Validate every schema
       run: |
         JSON_SCHEMAS=$(find ./jsonschema/google/events -name "*.json")
--- a/.github/workflows/validate_jsonschemas.yaml
+++ b/.github/workflows/validate_jsonschemas.yaml
@@ -26,4 +26,4 @@ jobs:
         node-version: '12'
     - name: Run Validate JSON Schema Script
       run: |
-        ./scripts/validat_jsonschema.sh
+        ./scripts/validate_jsonschema.sh
--- a/.github/workflows/validate_jsonschemas.yaml
+++ b/.github/workflows/validate_jsonschemas.yaml
@@ -1,0 +1,38 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Validate JSON Schemas
+on: [push, pull_request]
+jobs:
+  build:
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Check out repo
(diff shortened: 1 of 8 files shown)


.github/workflows/validate_jsonschemas.yaml
```diff
@@ -0,0 +1,38 @@
...
+  build:
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Check out repo
+      uses: actions/checkout@v2
+    - name: Setup Node for gen/postgen scripts
+      uses: actions/setup-node@v2
+      with:
+        node-version: '12'
+    - name: Install JSON schema schema
```
jskeet, line 27: My personal preference would be to put this logic into a shell script so it's easy to run manually... then run the shell script from the workflow. Not a huge deal though.

.github/workflows/validate_jsonschemas.yaml
```diff
@@ -0,0 +1,38 @@
...
+      run: |
+        curl -o jsonschema-schema.json http://json-schema.org/draft-04/schema
+    - name: Validate every schema
+      run: |
+        JSON_SCHEMAS=$(find ./jsonschema/google/events -name "*.json")
+        while IFS= read -r line ; do
+          npx ajv-cli@v3.0.0 validate -s jsonschema-schema.json -d $line;
+        done <<< "$JSON_SCHEMAS"
+    - name: Cleanup
```
jskeet, line 36: Do we particularly need to clean this up? Can't we regard it as a build artifact like the download of protoc?
Maybe put it in a `tmp` directory to make that clearer?

jskeet (approved):
A couple of suggestions, but I'm absolutely fine with this being merged as-is.

.github/workflows/validate_jsonschemas.yaml
grant: Moved script to `tmp`, so it's ignored naturally. Thanks!

.gitignore
```diff
@@ -7,3 +7,6 @@ tmp/
 
 # Local launch settings for .NET apps
 launchSettings.json
+
```
jskeet, line 10: Shouldn't need this any more as it's covered by `tmp/` on line 3, unless I'm mistaken about what the script does.

scripts/validate_jsonschema.sh
```diff
@@ -0,0 +1,27 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Validates the JSON Schemas are valid.
+# Must be run from the repo's root folder.
+
+# Install JSON schema schema
+mkdir tmp
```
jskeet, line 20: Possibly `rm -rf tmp` so that it's fine to run this more than once (or run this after running something else that creates tmp.)

jskeet (approved):
Two small suggestions, but otherwise great - thanks!

.gitignore
grant: Oops, yes.

scripts/validate_jsonschema.sh
grant: I've added `-p` so there's no warning that the folder exists.

It is already fine running this more than once, the curl command replaces whatever may be there.

Issue #195, opened 2021-03-26
Fix Protoc Plugin Pinned Version


grant:
## Expected Behavior

The chrusty tool that generates JSON schemas should be pinned to ensure reproducable results.

## Actual Behavior

It looks like the chrusty tool is using `master` from the repo.

## Steps to Reproduce the Problem

Run `./scripts/gen.sh`.

PR #196, opened 2021-03-26
fix: correctly pin json schema generator tool

grant:
This change correctly pins the golang protoc plugin (https://github.com/chrusty/protoc-gen-jsonschema) and runs the generator.

Fixes https://github.com/googleapis/google-cloudevents/issues/195

---

There have been some changes to that repo such that we don't want to pin to master right now.

https://github.com/chrusty/protoc-gen-jsonschema/tags

(Specificially with versions 0.9.8 and 0.9.9)

The diff undos the 0.9.8 changes (`binaryEncoding`and `format: binary` are invalid key/values that aren't used, that we don't really care about anyways)


jsonschema/google/events/cloud/audit/v1/LogEntryData.json
```diff
@@ -107,9 +107,7 @@
                   },
                   "value": {
                     "type": "string",
-                    "description": "Must be a valid serialized protocol buffer of the above specified type.",
-                    "format": "binary",
-                    "binaryEncoding": "base64"
```
grayside, line 112: Why drop this detail?

jskeet: This is interesting... this is precisely the problem I was running into when I was investigating running the generator on Windows. I'd thought that I was running an old version when it did this, and had fixed it by purging some things... but maybe it was the way round?

grant: There's a couple reasons we see this change:
- The previous generation didn't correctly pin the generator.
- This is because the generation was with this commit:
  - https://github.com/chrusty/protoc-gen-jsonschema/pull/51
  - This field is JSON Schema draft-4, not draft-7. We're using draft-07 (latest). This field is invalid with JSON schema and should not be included. The proper field is "contentEncoding".

The source of this issue can be traced to:
- https://github.com/alecthomas/jsonschema/issues/48
Which then can update this direct dependency we use:
- https://github.com/chrusty/protoc-gen-jsonschema

You can verify that this property is not in JSON schema draft 7:
- https://json-schema.org/draft-07/json-schema-release-notes.html

---

This binaryEncoding base64 detail in the schema is a recently introduced (harmless) bug (the field may look useful at the surface, but it is not a correct property and is not used by JSON schema).

---

We will need multiple changes to add base64 types in clients in the future, so it's infeasible to roll forward with a change. I'm requesting we properly pin the generated schema and investigate supporting base64 in a useful way tangentially.

jskeet (approved):
Somewhat-reluctantly approving - the change here will make the generated Java code (at least) worse. I definitely understand the desire to get the pinning right, but I'm nervous about the regression...

grant:
> Somewhat-reluctantly approving - the change here will make the generated Java code (at least) worse. I definitely understand the desire to get the pinning right, but I'm nervous about the regression...

Can you tell me exactly what is worse and why? Are the "format" or "binaryEncoding" fields used in the Java generator? Please link or file an issue so we can track this.

Perhaps we need multiple JSON schemas depending on the language.

jsonschema/google/events/cloud/audit/v1/LogEntryData.json
jskeet: I *thought* base64 string fields were being generated as `byte[]` properties in Java. I could have sworn they were at one point. I can't find evidence of that now though (which makes all of these fields somewhat unpleasant to use, unfortunately). So yes, if it's not actually being used to generate more pleasant clients, this is fine. (I think we should address this before going GA with Java though, and possibly Go too.)

grant: Base64 strings have always been represented as strings so far.

We could convert to a ["com.google.protobuf.ByteString"](https://github.com/googleapis/java-pubsub/blob/134273dc751e3bfb23c77384c8173c7b693a9757/proto-google-cloud-pubsub-v1/src/main/java/com/google/pubsub/v1/PubsubMessage.java#L178), but that would be a feature beyond what is normally generated.

grayside: Good to know we're on Draft 7. When this started chrusty was explicitly Draft 4.

grayside: To confirm then, the client libraries aren't using these properties for anything?

Since we don't provide guarantees on the schemas themselves, with that confirmed I'm good to LGTM

grayside: Confirmed!

PR #215, opened 2021-04-29
chore: init release please

grant:
Enables release please on this repo to enable simple releasing.

### DIFF
--- a/.github/release-please.yml
+++ b/.github/release-please.yml
@@ -1,0 +1,2 @@
+releaseType: simple
+handleGHRelease: true
--- a/.github/release-please.yml
+++ b/.github/release-please.yml
@@ -1,2 +1,16 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 releaseType: simple
 handleGHRelease: true


grant:
@jskeet Can you comment on this? I think it would be a good idea to start basic tags on this repository for us to better comment as to points in time of the source code.

https://github.com/googleapis/google-cloudevents/tags

jskeet:
Will have a look tomorrow. (UK public holiday today.)

jskeet:
I'm not sure what the tags or releases would be here... would it just be a minor version bump every time something new is added, and a patch bump for comment-only changes? What triggers an actual release?

Is the point of tagging just to make something slightly easier to refer to than a commit hash (that already provides point-in-time snapshots)?

I'm not asking these questions to be difficult or negative - I'm hoping that they'll help us (together) think about what we're trying to achieve in terms of tags/releases/docs etc.

In the meantime I'll approve this PR as I can see it unblocking experiments with trying different releases, and I can't see it doing actual harm.

jskeet (approved):

grant:
> I'm not sure what the tags or releases would be here... would it just be a minor version bump every time something new is added, and a patch bump for comment-only changes? What triggers an actual release?

The Release Please tool used here uses the commit messages of recent commits to trigger an actual GitHub release.

> Is the point of tagging just to make something slightly easier to refer to than a commit hash (that already provides point-in-time snapshots)?

There are many benefits of using GitHub's release tool and using Release Please for versioning and releasing. Git commit hashes provide a point in time that doesn't necessarily reflect the git commit messages. Additionally, individual git commits, like the 151+ here, are not as easy to organize automatically without releases.

> I'm not asking these questions to be difficult or negative - I'm hoping that they'll help us (together) think about what we're trying to achieve in terms of tags/releases/docs etc.

Understood. For example, I'm trying to use the Go unmarshalling library in our samples and I'm blocked by having releases.

> In the meantime I'll approve this PR as I can see it unblocking experiments with trying different releases, and I can't see it doing actual harm.

Thanks. If we find out there is harm or pain, rather than expected, I'm happy to revert or modify the GitHub release process.

jskeet:
> For example, I'm trying to use the Go unmarshalling library in our samples and I'm blocked by having releases.

Do you mean you're blocked by the lack of releases from *this* repo, or by the lack of releases in the Go repo? I completely see how it makes sense to have releases for client libraries - I'm a little less convinced about the need in *this* repo, but I don't think it'll do any harm, and it may make us think about things we otherwise wouldn't (such as whether we should only generate libraries from tagged releases, for example).

grant:
> > For example, I'm trying to use the Go unmarshalling library in our samples and I'm blocked by having releases.
> 
> Do you mean you're blocked by the lack of releases from _this_ repo, or by the lack of releases in the Go repo? I completely see how it makes sense to have releases for client libraries - I'm a little less convinced about the need in _this_ repo, but I don't think it'll do any harm, and it may make us think about things we otherwise wouldn't (such as whether we should only generate libraries from tagged releases, for example).

It's both repos really.

It will be a lot easier to release our client libraries with versions if we can generate the libraries using a cut version from the source in this repo.

grant:
I'm not trying to quickly force/merge things through, but I really need to finish using the libraries in Eventarc samples and move on to other projects.

PR #220, opened 2021-05-21
ci: adds autogen node script

grant:
Per the unmarshal autogen plan, this GitHub workflow runs the generator for the Node client and generates a PR for that Node repo if there is a non-0 diff whenever we push an update to the default branch of this repo.

Please treat the versioning question separate from this PR.


grayside (approved):
LGTM but perhaps @averikitsch or someone on the GitHub Automation team should be tagged for review?

grant:
> LGTM but perhaps @averikitsch or someone on the GitHub Automation team should be tagged for review?

Sure, Averi can you take a look at this PR and lmk if there are any issues?

.github/workflows/client_node.yaml
```diff
@@ -0,0 +1,63 @@
...
+name: Run Node Client Generator and Create Pull Request
+on:
+  push:
+    branches:
+      - "master"
+jobs:
+  create_pr:
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Checkout this repo
```
averikitsch, line 24: Do you need to checkout this repo?

.github/workflows/client_node.yaml
```diff
@@ -0,0 +1,63 @@
...
+        if [[ `git status --porcelain` ]]; then
+          git config --global user.email "yoshi-code-bot@google.com"
+          git config --global user.name "Google Bot"
+          git checkout -b "pr_bot-$GITHUB_RUN_ID"
+          git add -A
+          NOW=$(date +"%m-%d-%Y")
+          git commit --message "feat: run generator (${NOW})"
+          git push origin "pr_bot-$GITHUB_RUN_ID"
+          echo "Created commit: $(git rev-parse HEAD)"
+          gh pr create \
```
averikitsch, line 57: nit: there's an action you can use, but this is just as clean

averikitsch (approved):
Just nits

.github/workflows/client_node.yaml
grant: Yes, this repo contains schema files used for generation.

.github/workflows/client_node.yaml
grant: Thanks. Using the CLI is easier for manual testing, so let's keep as is.

PR #226, opened 2021-06-09
Prototype of protobuf annotations to describe extension attributes

jskeet:
See https://github.com/googleapis/google-cloudevents-dotnet/pull/62 for an example of what we can do with this from a library perspective, as well as it just being useful in general.

The precise details of the annotations can certainly be changed - this is more of a sketch than anything else.
Note that when we decide we want to do this, we'll need to make the change internally first so that it will then be propagated.


jskeet:
Closed, replaced by #228

PR #235, opened 2021-06-30
build: update renovate config and quicktype wrapper deps

JustinBeckwith:
The current renovate config has some problems:
- It's not configured to do lock file updates
- It's configured to run once every 6 months

This opens us up to dependabot vulnerability alerts, of which we currently have a few.  This PR:
- Attempts to fix the renovate config
- Moves from pinned dependencies in tools/quicktype-wrapper to floating dependencies (like we do everywhere else)
- Adds ignore files required for `gts` to do formatting
- Runs `npm run fix`, applying a few style changes

I noticed during this exercise that there are no tests or CI specifically for this directory, so it's very difficult to know if I'm breaking something.


renovate.json
```diff
@@ -2,7 +2,10 @@
   "extends": [
     "config:base"
   ],
-  "schedule": [
-    "every 6 months"
-  ]
```
grant, line 7: We do want to have a schedule of some sort and not just bumping patch versions of dependencies. That was the original reason for renovate here.

Can we have some sort of schedule to prevent too many trivial PRs?

grant (approved):
Thanks. These changes look good to me. We definitely can improve the tooling and testing with this part of the generation process, I can tinker with some part of this repo as I make changes.

renovate.json
JustinBeckwith: If we move towards using the `^` in versions here (as I started to do),  you'll only get renovate PRs for semver major releases, and a weekly lock file update.  Those tend to be fairly infrequent.

grant:
@JustinBeckwith Is there any way to prevent trivial PRs like https://github.com/googleapis/google-cloudevents/pull/236 ? Or is this PR expected?

JustinBeckwith:
In the specific case of TypeScript, they like to make breaking changes in minor releases - which is why I typically put the `~` on it.  For specific dependencies, you may choose to create different rules in the renovate config - or you can roll the dice and use a `^` :)   Generally here though - we want to keep dependencies up to date.  Yes, these can be a minor inconvenience, but they take seconds to approve and merge in most cases, assuming CI is set up on the project.  The alternative is reactively fixing dependabot alerts as they come up.

PR #253, opened 2021-07-09
Declare the protos on the default branch to be stable

jskeet:
This stability is restricted to just the protos, and only those on
the default branch.

cc @hrasadi and @haolicopter (who should probably have more rights on this repo)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,12 +11,11 @@ The "source of truth" for the event data format is in the form of
 See [proto/README.md](proto/README.md) for more
 information about the messages used to describe CloudEvents.
 
-## Stability warning
+## Stability
 
-The entire content of this repository should be regarded as highly
-unstable until this warning is removed. The schema may change, the
-CloudEvent "type" attribute values may change, the protobuf message
-may change. Even the core `cloud_event_type` annotation may change.
+The protobuf files in the [proto](proto/) directory should be
+regarded as stable with no further breaking changes. Other files
+such as generated JSON schema should not be regarded as stable.
 
 ## CloudEvents in this repository
 


README.md
```diff
@@ -11,12 +11,17 @@ The "source of truth" for the event data format is in the form of
...
-## Stability warning
+## Stability
 
-The entire content of this repository should be regarded as highly
-unstable until this warning is removed. The schema may change, the
-CloudEvent "type" attribute values may change, the protobuf message
-may change. Even the core `cloud_event_type` annotation may change.
+The protobuf files in the [proto](proto/) directory of the default
+branch should be regarded as stable, with no further breaking
+changes. New events should not be added to the default branch until
```
grant, line 18: I don't think have or will use multiple branches (without intention for merging to the default branch) for new events in the future. Is this a new proposal? Do we plan for alpha beta event versioning for new events or go from branch to stable?

Describing this process a bit more might be useful - probably in a different README file.

jskeet: It's more leaving the door open to that in the future, rather than making a concrete proposal right now. I could take it out if you'd prefer.

grant: My opinion:

We should keep this top portion to the point. Viewers are reading this repo from Eventarc and GCF (GA) docs.

Suggestion, instead of a section, just have this paragraph with this formatting:

> **Stability**: The protobuf files in the [proto](proto/) directory should be regarded as stable with no further breaking changes. Other files such as generated JSON schema should not be regarded as stable.

We don't need to be extra verbose here. Developers want to see CloudEvents in this repo. We can add further details for maintainers and the exact meaning of stable in a different README (possibly link it).

We also have multiple GA products like Eventarc, GCF .NET, etc, that use these CloudEvents. It's a little scary seeing big disclaimers in the beginning of a repo.

grant (changes requested):

jskeet:
I would definitely prefer it to still be a section - I think it should be highlighted, both in terms of what folks can and can't rely on. I think a section does that better than just a single bold word in a paragraph.

I'd be happy with your suggested wording but keeping it as a single-paragraph section. I'd still like to put *something* about new events, but that could be in the protos/README.md file.

(I also don't want to assume where readers are coming from. Some may have arrived on the page via the Eventarc and GCF docs - others could be arriving from all manner of other places. I also think that if developers reach active repository docs, they should expect there to be more detail than in the product docs - it needs to serve both readers and contributors.)

grant:
Okay. I don't see commonalities with a stability section with other Google API / client repos, and this change is generally good. Some middle ground would be good as you suggested. I'll approve  🙂

grant (approved):

jskeet:
> Okay. I don't see commonalities with a stability section with other Google API / client repos

That's because the content of those repos is normally consumed via published libraries, which are versioned with semver to make it obvious. The main content of this repo is really the protos themselves (from my standpoint, anyway).

Arguably we really *should* have a stability notice on googleapis/googleapis, but at the moment it's slightly unclear what that would actually be. That should get better over time though...

> Some middle ground would be good as you suggested. I'll approve 

Thanks. Will change the wording as discussed, force push, and rebase.

Issue #250, opened 2021-07-06
JSON schema google.protobuf.Duration Should Be Treated as string


grant:
It looks like the JSON serialization of a `google.protobuf.Duration` is a string like `"660s"`. Currently we misrepresent the JSON schema as an object. This affects types like Cloud Build objects.

- Example JSON: https://github.com/googleapis/google-cloudevents/blob/master/testdata/google/events/cloud/cloudbuild/v1/BuildEventData-simple.json
- Proto: https://github.com/googleapis/google-cloudevents/blob/master/proto/google/events/cloud/cloudbuild/v1/data.proto
  - `queue_ttl` is a `google.protobuf.Duration` (from `google/protobuf/duration.proto`)

I've filed an upstream issue here: https://github.com/chrusty/protoc-gen-jsonschema/issues/63

PR #254, opened 2021-07-11
feat: improve jsonschema types for protobuf well known types

grant:
Fixes https://github.com/googleapis/google-cloudevents/issues/250

Uses `PROTOC_PLUGIN_VERSION=1.0.0`.

- Better JSON representation for `google.protobuf.Value`
- Better JSON representation for `google.protobuf.Duration`

This should result in better parsing for these well known types, and types that aren't always JSON objects (hence the oneOf).

See: https://developers.google.com/protocol-buffers/docs/reference/google.protobuf
- e.g. https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Value

Gen'd updates to language clients would follow this PR. Unfortunately the `oneOf` syntax isn't the greatest with quicktype. Although it is the most accurate for the JSON (not everything is an object), and easier to deal with downstream. Look at the protobuf docs for your own assessment.

### DIFF
--- a/jsonschema/google/events/cloud/cloudbuild/v1/BuildEventData.json
+++ b/jsonschema/google/events/cloud/cloudbuild/v1/BuildEventData.json
@@ -138,18 +138,8 @@
             "description": "Stores timing information for pulling this build step's\n builder image only."
           },
           "timeout": {
-            "properties": {
-              "seconds": {
-                "type": "integer",
-                "description": "Signed seconds of the span of time. Must be from -315,576,000,000\n to +315,576,000,000 inclusive. Note: these bounds are computed from:\n 60 sec/min * 60 min/hr * 24 hr/day * 365.25 days/year * 10000 years"
-              },
-              "nanos": {
-                "type": "integer",
-                "description": "Signed fractions of a second at nanosecond resolution of the span\n of time. Durations less than one second are represented with a 0\n `seconds` field and a positive or negative `nanos` field. For durations\n of one second or more, a non-zero value for the `nanos` field must be\n of the same sign as the `seconds` field. Must be from -999,999,999\n to +999,999,999 inclusive."
-              }
-            },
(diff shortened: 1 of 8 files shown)


jsonschema/google/events/cloud/cloudbuild/v1/BuildEventData.json
```diff
@@ -265,18 +257,8 @@
...
-          "description": "Signed seconds of the span of time. Must be from -315,576,000,000\n to +315,576,000,000 inclusive. Note: these bounds are computed from:\n 60 sec/min * 60 min/hr * 24 hr/day * 365.25 days/year * 10000 years"
-        },
-        "nanos": {
-          "type": "integer",
-          "description": "Signed fractions of a second at nanosecond resolution of the span\n of time. Durations less than one second are represented with a 0\n `seconds` field and a positive or negative `nanos` field. For durations\n of one second or more, a non-zero value for the `nanos` field must be\n of the same sign as the `seconds` field. Must be from -999,999,999\n to +999,999,999 inclusive."
-        }
-      },
       "additionalProperties": true,
-      "type": "object",
+      "type": "string",
```
jskeet, line 261: I'm slightly surprised there's no format part here - does this mean there's no way of a JSON schema consumer telling the difference between a duration and a string?

jskeet (commented):
Happy with bytes and Value - I think Duration still needs some work though.

jskeet (approved):
(Approving as it's definitely better than before.)

jsonschema/google/events/cloud/cloudbuild/v1/BuildEventData.json
grant: We currently don't have it but we can add the formatting in the future.

Filed https://github.com/chrusty/protoc-gen-jsonschema/issues/66

PR #305, opened 2021-12-11
Exclude copybara branch from the Generate workflow

marcgovi:
The `generate` workflow runs for any `push` or any `pull request` action and makes sure no changes have been introduced to the proto schemas.
The problem is that `copybara` branch might in fact be performing changes to those schemas (eg: adding a new one), thus making it fail, and not allowing admins to merge it without disabling branch protection first.

I'm proposing to skip this workflow for the `copybara` branch to reduce manual intervention.
FYI, once a `copybara` PR is submitted to `main`, the generator will run anyway and create a PR with any new changes.

### DIFF
--- a/.github/workflows/generate.yaml
+++ b/.github/workflows/generate.yaml
@@ -13,9 +13,15 @@
 # limitations under the License.
 
 name: Generate
-on: [push, pull_request]
+on:
+  push:
+    branches-ignore:
+    - copybara
+  pull_request:
 jobs:
   build:
+    # Skip when it is a pull request from copybara as there might be changes in the proto schemas
+    if: github.head_ref != 'copybara'
     runs-on: ubuntu-18.04
     steps:
     - name: Check out repo


grayside:
My understanding is this diff check is in place primarily to make sure humans are paying attention to potential compatibility breaks. Is a possibly better path to only error if there are deletions or changes but not on pure additions?

@grant do you have insight on the role of the diff you could share?

grayside (changes requested):
Blocking merge to make sure (minimally) I understand the risks of this change.

grant (approved):
Great, this is an improvement. See some background below:

Happy to add perspective.

Re Adam:

> My understanding is this diff check is in place primarily to make sure humans are paying attention to potential compatibility breaks. Is a possibly better path to only error if there are deletions or changes but not on pure additions?
> 
> @grant do you have insight on the role of the diff you could share?

This GitHub workflow was mainly to add a ✅ check to ensure that any generated resources were created along with the pull request change (for example, changing the JSON schema generator / catalog generator / README generator).

Without this check, there were additional PRs to sync `main`, making me send lots of PR.

The `copybara` branch is special compared to all other branches, where it is not expected to have generated resources, so this PR lgtm.

---

For testing proto incompatibilities, we don't have an exact test that checks API-level incompatibilities.

The best related thing we have for that is testing the sample JSON payloads which test against our protos:

https://github.com/googleapis/google-cloudevents/tree/main/testdata

Maybe there's a proto diff we could add - that should probably be a g3 presubmit error though and not downstream here (we can't do much at that point).

Note: Make sure to check with Adam before going with this PR.

grayside:
> Maybe there's a proto diff we could add - that should probably be a g3 presubmit error though and not downstream here (we can't do much at that point).

Definitely, I see it as a defense in-depth discussion, but also making sure I understand the pieces in play.

grayside (approved):

PR #345, opened 2022-03-31
Fetch CloudEvent type only for matching payload types

tommyreddad:
If we don't do this, then all JSON schemas get associated with every event type defined in events.proto. But those events might not all have the same payload.

For example before this change, with the following events defined in the same events.proto file:
```proto
message FooEvent {
  option (google.events.cloud_event_type) = "google.example.foo.v1.blah";
  FooEventData data = 1;
}

message BarEvent {
  option (google.events.cloud_event_type) = "google.example.bar.v1.blah";
  BarEventData data = 1;
}
```

The schemas would look like:
```json
{
  "name": "FooEventData",
  "cloudeventTypes": [
    "google.example.bar.v1.blah",
    "google.example.foo.v1.blah"
  ]
}
```
```json
{
  "name": "BarEventData",
  "cloudeventTypes": [
    "google.example.bar.v1.blah",
    "google.example.foo.v1.blah"
  ]
}
```

This change makes sure that the event types listed in the JSON schema are correct. After this change, the schemas would look like:
```json
{
  "name": "FooEventData",
  "cloudeventTypes": [
    "google.example.foo.v1.blah"
  ]
}
```
```json
{
  "name": "BarEventData",
  "cloudeventTypes": [
    "google.example.bar.v1.blah",
  ]
}
```

### DIFF
--- a/tools/proto2jsonschema/postgen.js
+++ b/tools/proto2jsonschema/postgen.js
@@ -21,12 +21,12 @@ const protobufjs = require('protobufjs');
 const flatten = require('flat');
 
 /**
-* This tool polishes the JSON schemas with a few modifications:
-* - Adds "$id" – "." delimited ID – e.g. "google.events.cloud.audit.v1.LogEntryData"
-* - Adds "name" – The name for the JSON schema – e.g. "LogEntryData"
-* - Adds "examples" - A list of paths to the test event data associated with the schema
-*   - e.g. ["https://googleapis.github.io/google-cloudevents/testdata/google/events/cloud/audit/v1/LogEntryData-pubsubCreateTopic.json"]
-*/
+ * This tool polishes the JSON schemas with a few modifications:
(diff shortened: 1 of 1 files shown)


grayside:
Do you know of an example of this in the existing event types we can use to test the change?

tommyreddad:
> Do you know of an example of this in the existing event types we can use to test the change?

There is none, I discovered the issue as I was developing a new event type. It works in my manual tests.

grayside (approved):
I read through the code and it seems reasonable & greenlit the tests. Those passed, and Generate/build shows no diff, which means the existing event types that shouldn't be affected by this change aren't.

With an event type to help us exercise the tests, it's difficult to know if this will be stable. I think we should merge and keep in mind when that event type arrives more work may be needed.

tommyreddad:
Thanks! I don't think I can merge this PR though:

>Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests.

PR #350, opened 2022-04-30
Fixing minor typos in comments

TheIronDev:
Fixing minor typos in the comments.


averikitsch:
These may be copied from internal code. I will follow up.

marcgovi:
Yes that's right @averikitsch, this schema is pushed from internal code. For simplicity I would suggest to change it there.

averikitsch (approved):

PR #380, opened 2022-08-01
chore: Use .NET 6 instead of .NET Core 3.1 for .NET validation

jskeet:
This should fix the .NET failure in #378.

### DIFF
--- a/.github/workflows/validate_clients.yaml
+++ b/.github/workflows/validate_clients.yaml
@@ -30,9 +30,9 @@ jobs:
       with:
         repository: googleapis/google-cloudevents-dotnet
         path: google-cloudevents-dotnet
-    - name: Setup .NET Core 3.1
+    - name: Setup .NET 6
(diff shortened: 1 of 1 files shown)


ace-n (approved):

ace-n:
@jskeet can you address the `Verify No Changes` check?

Thanks! 🙂

jskeet:
Not at the moment I'm afraid - I'm on holiday. I'll look on Monday.

jskeet:
@ace-n: "Verify no changes" fails because I believe it runs the generator - which would create changes. That would be fixed by merging #377, but that fails because of the .NET version issue addressed by this PR.

So basically I think we need to merge this change, at which point we can rebase #377 and it will be clean (and other PRs will be fine too). Let me know if you're happy for me to merge this with the failure still present.

(We could *potentially* have a single PR containing both the generator result and the fix here, which would be green, but they're really different aspects...)

PR #430, opened 2022-10-10
chore: Add link to triggering pr

averikitsch:


.github/workflows/generate_jsonschema.yaml
```diff
@@ -36,7 +36,14 @@ jobs:
           command: pr
           upstream_owner: ${{ github.repository_owner }}
           upstream_repo: "google-cloudevents"
-          description: "Generated in GitHub action: https://github.com/${{ github.repository }}/actions/workflows/generate_jsonschema.yaml"
+          description: |
+              :robot: Auto-generated Pull Request.
+              Triggered by [${{ github.event.repository.full_name }}@${{ github.event.head_commit.id }}](https://github.com/${{ github.event.repository.full_name }}/commit/${{ github.event.head_commit.id }})
+
+              * **Generator configuration:** https://github.com/${{ github.repository }}/blob/${{ github.sha }}/.github/workflows/generate_jsonschema.yaml
```
muncus, line 43: I'm not sure we need the SHAs in there, but i could go either way.
On the one hand, its nice to see the config that triggered the generation. On the other, i dont expect these to change that much. Do you?

muncus (approved):

.github/workflows/generate_jsonschema.yaml
averikitsch: I don't expect it to change a lot, but it might be nice to track in case we ever need to go back through PRs.

PR #431, opened 2022-10-12
chore: add new test data

averikitsch:


muncus (approved):
With the failure of the "unexpected change" detector, does this need to be force-submitted? seems like a bad flow for testdata.

should test data trigger regeneration of docs?

averikitsch:
> With the failure of the "unexpected change" detector, does this need to be force-submitted? seems like a bad flow for testdata.
> 
> should test data trigger regeneration of docs?

./scripts/gen.sh script bundles the generation of the jsonschemas and associated docs. The jsonschema links to these examples. We can either re evaluate those decisions or we can make the test optional

PR #461, opened 2022-12-19
chore: add go_package to third_party/.../monitored_resource

grayside:
This change allows using monitoring_resource as a proto dependency in Go without customizing Go import path mappings in the build process.


grayside:
Unexpected changes failure seems unrelated to this PR.

engelke (approved):

PR #477, opened 2023-01-17
ci: Add Java Smoke Tests

averikitsch:


.github/workflows/test_clients.yaml
```diff
@@ -29,18 +29,54 @@ jobs:
...
+      DATA_SOURCE_CHECKOUT_PATH: google-cloudevents
+    steps:
+      - name: Java Library > Checkout Repository
+        uses: actions/checkout@v3
+        with:
+          repository: googleapis/google-cloudevents-java
+          path: ${{ env.LIBRARY_CHECKOUT_PATH }}
+          ref: snowpea-build
+
+      - name: Proto Schemas > Checkout Repository
```
grayside, line 62: I wonder if we could do a single checkout of proto schemas and reuse across all languages?

averikitsch: I don't think it's worth the work trying to create a parent job to checkout and have child jobs uses that one source. Are you concerned or just trying to minimize each language setup?

.github/workflows/test_clients.yaml
```diff
@@ -29,18 +29,53 @@ jobs:
...
+    env:
+      LIBRARY_CHECKOUT_PATH: library
+      DATA_SOURCE_CHECKOUT_PATH: google-cloudevents
+    steps:
+      - name: Java Library > Checkout Repository
+        uses: actions/checkout@v3
+        with:
+          repository: googleapis/google-cloudevents-java
+          path: ${{ env.LIBRARY_CHECKOUT_PATH }}
+          ref: snowpea-build
```
averikitsch, line 60: Note to self: remove before merging

.github/workflows/test_clients.yaml
grayside: Just thinking about avoiding redundant actions and per-language complexity. But agreed, refactoring for that is a good future improvement.

grayside (approved):

PR #482, opened 2023-01-30
testing: add go to test-client workflow

grayside:
* Adds Go to test client workflow
* Limits workflow permission to `contents: read` for .NET, Java, and Go

### DIFF
--- a/.github/workflows/test_clients.yaml
+++ b/.github/workflows/test_clients.yaml
@@ -108,6 +108,7 @@ jobs:
         go-version: 1.19.x
 
     - name: Set Environment
+      # Validation Test execution requires absolute path to testdata.
       run: echo "GENERATE_DATA_SOURCE=${PWD}/${DATA_SOURCE_CHECKOUT_PATH}" | tee -a $GITHUB_ENV
 
     - name: Install the generator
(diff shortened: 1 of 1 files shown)


.github/workflows/test_clients.yaml
```diff
@@ -78,3 +80,44 @@ jobs:
...
+      with:
+        path: ${{ env.DATA_SOURCE_CHECKOUT_PATH }}
+
+    - name: Setup Go
+      uses: actions/setup-go@v3
+      with:
+        go-version: 1.19.x
+
+    - name: Set Environment
+      run: echo "GENERATE_DATA_SOURCE=${PWD}/${DATA_SOURCE_CHECKOUT_PATH}" | tee -a $GITHUB_ENV
```
averikitsch, line 111: What is PWD referring to?

grayside: It's a standard variable that is loaded with the current working directory. The validation tests execution require this variable to be an absolute path. Feedback from @muncus in [google-cloudevents-go ](https://github.com/googleapis/google-cloudevents-go/pull/151) was to avoid the extra complication of `realpath` and use simpler string concatenation.

PR #502, opened 2023-04-14
docs: First pass at moving the Pub/Sub binding into this repo.

jskeet:
Original source:
https://github.com/google/knative-gcp/blob/main/docs/spec/pubsub-protocol-binding.md

I've modified the text in a few places as the CloudEvents spec wording has been tightened up since that doc was written:

- Event data isn't an attribute
- There are now three content modes: binary, structured and batch. (We could easily extend this spec to support batch, but we haven't done so yet.)
- The context attributes defined in the CloudEvents spec now have a name: "core context attributes"

Reading through this, it looks like we don't have a convention for extension attributes, and we don't have any examples of them either. We can approach that in a follow-up PR.

cc @ymalik


hrasadi (approved):

marcgovi (approved):

jskeet:
> Given docs/ are usually repository docs, and this is actually a CloudEvent spec, perhaps this should be in a spec/ directory?

Happy to do that if we'd prefer - will discuss internally (and add you to that conversation).
(I think I'd prefer docs because it's effectively documentation and I can see there being more docs, but I suspect in a spec directory this might always be the only document...)

marcgovi (approved):

